PR #1, opened 2024-08-07
optimizing Blob cursor for touch mobile devices

M7zain:
Moved event listeners for mousemove and touchmove into a new useEffect hook, attaching them directly to the main element to ensure touch events are handled smoothly from the start.


DavidHDev:
Hi @M7zain , i've already added this in a commit yesterday, along with a few other improvements so I will close the PR:
https://github.com/DavidHDev/react-bits/commit/029555c537e4a6548aca40006f7cd5d8f8d42b8c

PR #25, opened 2025-01-16
feat(cli): Dual built registries

ieedan:
In the last week I have realized a few things from looking at this project and others and that drove me to make some improvements...

First of all I have changed the way you supply repos to init to make it more concise:
```bash
# new
jsrepo init github/ieedan/std

# old
jsrepo init --project --repos github/ieedan/std
```

Next I have added support for registries hosted on your own domain. This allows for more customization and is really helpful if you are doing something like in this project where you have 2 component sets with different code.

This PR proposes a new way of building with jsrepo that goes to the public directory. You would have 2 registries:
```bash
https://reactbits.dev/default
https://reactbits.dev/tailwind
```

This way users can specify from the beginning which component set they want to use. 

## Changes on your end
- To build the registry properly with this setup you need to change your vercel configuration and override the `build` script. (Instead of `vite build` it should be `npm run build`)

## Testing the registry
Since you can't test the registry using git branches anymore you instead can just test it locally:
```bash
https://localhost:5173/default
```

## Other changes
- renamed components prefixed with `Tailwind` to remove the prefix since it is no longer necessary

Let me know what you think!


DavidHDev:
Hi @ieedan , main problem is, the last thing i want is to have two routes in my app (default/tailwind), so I'm not sure this is the correct approach

ieedan:
Why is that?

DavidHDev:
Maybe I didn't understand this right. You're not proposing any changes in the way the routing of the app currently works, is that right? The only things changing would be the cli commands

ieedan:
Correct the routes that you see are static routes under the `./public` directory. Your routing doesn't need to change unless you have conflicting routes in the future. 

Those routes will be generated by the cli when vercel runs `npm run build`

DavidHDev:
thanks!

ieedan:
If you haven't already noticed this will be breaking to anyone using the old jsrepo github/davidhdev/react-bits route up to you how we should handle that!

DavidHDev:
That's not an issue, but I do hope we can keep this change as a stable solution?

ieedan:
Hard to ever say but I think this is a better solution that what we had before!

DavidHDev:
Alright, sounds good, thanks again

ieedan:
Ah deployment is failing because you forgot to change your build config I believe...

image

Just change your build command to `npm run build` instead of `vite build`

DavidHDev:
Are you sure it's not supposed to be done like this in package.json?
Screenshot 2025-01-17 at 15 19 38
Screenshot 2025-01-17 at 15 20 20

ieedan:
Ah yes i just noticed this now I will open a PR rn

ieedan:
Sorry about that! I must have forgot to test that command specifically.

DavidHDev:
No worries!

ieedan:
You should notice when you add from the url it should be much faster since you aren't having to do any of the url resolution for github anymore.

DavidHDev:
I'll play around with it for sure! Thanks and great work!

PR #27, opened 2025-01-17
fix: Fix build script

ieedan:
That's embarrassing!


DavidHDev:
Looks like it passed, cool!

PR #37, opened 2025-01-27
Added Typescript support ShapeBlur

saj13h:


DavidHDev:
@saj13h Thanks for this! Before I merge it, did you make sure to test these TS components? Just asking for peace of mind 💯

src/constants/code/Backgrounds/shapeBlurCode.js
```diff
@@ -2,6 +2,8 @@ export const shapeBlur = {
   installation: `npm i three`,
   cliDefault: `npx jsrepo add https://reactbits.dev/default/Backgrounds/ShapeBlur`,
   cliTailwind: `npx jsrepo add https://reactbits.dev/tailwind/Backgrounds/ShapeBlur`,
+  cliTsDefault: `npx jsrepo add https://reactbits.dev/ts-default/Backgrounds/ShapeBlur`,
```
DavidHDev, line 5: The URLs are incorrect, they should be:
  cliTsDefault: `npx jsrepo add https://reactbits.dev/ts/default/Backgrounds/ShapeBlur`,
  cliTsDefault: `npx jsrepo add https://reactbits.dev/ts/tailwind/Backgrounds/ShapeBlur`,

saj13h:
i tested and it works but there is a compile error showing. 

message: Could not find a declaration file for module 'three'. '/Users/saj13h/Desktop/nexttailwindtest/node_modules/three/build/three.module.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/three` if it exists or add a new declaration (.d.ts) file containing `declare module 'three';`

there is a quick fix for it but i don't know it is the way: 
is installing a package : npm i --save-dev @types/three

even without this package it works

DavidHDev:
@saj13h that should be fine, three has no TS support out of the box as far as I know, so there's not much we can do about it! Could you maybe add a comment at the top of the TS files to share this info?

DavidHDev:
Thanks for the contribution! 💯

PR #38, opened 2025-01-27
Removing problematic position property from Decrypt text tailwind component 

tonyantony300:
```
 <DecryptedText
                text="Secrets"
                animateOn="view"
                revealDirection="center"
              />
```
Doing this reserves space on the DOM but text not visible. I found this is the problem with `relative` positioning. Removing it fixes it.
Screenshot 2025-01-26 at 5 00 31 PM

### DIFF
--- a/src/constants/code/TextAnimations/decryptedTextCode.js
+++ b/src/constants/code/TextAnimations/decryptedTextCode.js
@@ -492,5 +492,5 @@ return (
     </span>
   </motion.span>
 )
-}`,
-};
+}`
+}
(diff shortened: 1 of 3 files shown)


DavidHDev:
Hey, you also need to update the code snippet in `decryptedTextCode.js` to reflect the changes for those who use copy-paste.

DavidHDev:
Also please don't add formatting changes in the PRs, I won't make you revert it but just so you know.

tonyantony300:
I think its done now. Let me know if anything needs to be changed.

DavidHDev:
yep, looks great! Thanks 🥇

PR #53, opened 2025-01-28
FEAT: blob-cursor typescript version1.0

SandunJay:


src/ts-default/Animations/BlobCursor/Blob-Cursor.tsx
DavidHDev, line 1: Component should respect the naming of other components, it should be `BlobCursor.tsx`

DavidHDev:
You also need to update:
- the `blobCursorCode.js` file with the new CLI command and new import, see other component files for an example
- the `BlobCursorDemo.jsx` file to pass in the new CLI command

SandunJay:
CLI command and the import was added

PR #54, opened 2025-01-28
feat: Add readmes and concurrent builds

ieedan:
Adds a script to generate readmes as mentioned in #52 let me know if the description should be different for each registry I wasn't super happy with the wording.

Also parallelizes the jsrepo build command and removes npx.
- Since we have jsrepo as a dependency npx is not necessary
- Now that we are building 4 registries its pretty slow to do them one by one

Also bumps jsrepo to 1.29.1 no changes here just bug fixes

> Also every time I have to PR the package.json it does this since there isn't any formatter setup maybe it would be good to add one?


DavidHDev (approved):

DavidHDev:
Looks good to me, thanks!

Regarding the formatting, maybe I'll add something in the near future, it's not that big of a concern for me personally since it's also possible that most people will probably format the code files to their own preferences anyway after installing the components

DavidHDev:
I get it, I'll think about it! Since it was mostly just me working on this until this point I wasn't considering it

DavidHDev:
Gonna merge this 🙌

PR #57, opened 2025-01-29
FEAT: TS TrueFocus

SandunJay:


DavidHDev:
Hey, since we already have the tailwind version, you could also add the ts/tailwind version as well, please update the PR to include it so we can have all 4 versions 👍

src/ts-default/TextAnimations/TrueFocus/TrueFocus.tsx
```diff
@@ -0,0 +1,122 @@
+import { useEffect, useRef, useState } from "react";
+import { motion } from "framer-motion";
+import "./TrueFocus.css";
+
+const TrueFocus = ({
+  sentence = "True Focus",
```
DavidHDev, line 6: You need to add an interface for the props, what's the point of making it a .tsx files if we don't do that?

src/ts-default/TextAnimations/TrueFocus/TrueFocus.tsx
```diff
@@ -0,0 +1,122 @@
+import { useEffect, useRef, useState } from "react";
+import { motion } from "framer-motion";
+import "./TrueFocus.css";
+
+const TrueFocus = ({
```
DavidHDev, line 5: this should be something like `const TrueFocus: React.FC<TrueFocusProps> = ({`

src/ts-default/TextAnimations/TrueFocus/TrueFocus.tsx
```diff
@@ -0,0 +1,122 @@
...
+  glowColor = "#ef4444",
+  animationDuration = 0.5,
+  pauseBetweenAnimations = 1,
+}) => {
+  const words = sentence.split(" ");
+  const [currentIndex, setCurrentIndex] = useState(0);
+  const [lastActiveIndex, setLastActiveIndex] = useState<number | null>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const wordRefs = useRef<(HTMLSpanElement | null)[]>([]);
+  const [focusRect, setFocusRect] = useState({ x: 0, y: 0, width: 0, height: 0 });
```
DavidHDev, line 19: this should be typed:

```javascript
  const [focusRect, setFocusRect] = useState<{ x: number; y: number; width: number; height: number }>({
    x: 0,
    y: 0,
    width: 0,
    height: 0,
  });
```

src/constants/code/TextAnimations/trueFocusCode.js
```diff
@@ -3,6 +3,7 @@ import { generateCliCommands } from '@/utils/utils';
 import code from '@content/TextAnimations/TrueFocus/TrueFocus.jsx?raw';
 import css from '@content/TextAnimations/TrueFocus/TrueFocus.css?raw';
 import tailwind from '@tailwind/TextAnimations/TrueFocus/TrueFocus.jsx?raw';
+import tsCode from '@ts-default/TextAnimations/TrueFocus/TrueFocus.tsx?raw';
 
 export const trueFocus = {
   ...(generateCliCommands('TextAnimations/TrueFocus', ['default', 'tailwind'])),
```
DavidHDev, line 9: After you add the tailwind/ts version, you can remove the array from this function call:
`...(generateCliCommands('TextAnimations/TrueFocus')),`

DavidHDev:
Implemented in:

https://github.com/DavidHDev/react-bits/pull/58

PR #58, opened 2025-01-29
FEAT: Add TrueFocus (TS)

ipuppyyt:


DavidHDev:
Hey, someone already opened a PR for this yesterday.. this code is better however, so I would propose going with this instead of waiting for that merge

DavidHDev:
great work thanks! 💯

PR #60, opened 2025-01-29
Added TypeScript version of "Dock" component under "Component" section

whitedevil-glitch:


src/constants/code/Components/dockCode.js
```diff
@@ -1,14 +1,18 @@
...
-import code from '@content/Components/Dock/Dock.jsx?raw';
-import css from '@content/Components/Dock/Dock.css?raw';
+import code from "@content/Components/Dock/Dock.jsx?raw"
+import css from "@content/Components/Dock/Dock.css?raw"
+import tsCode from "@ts-default/Components/Dock/Dock.tsx?raw"
+import tsTailwind from "@ts-tailwind/Components/Dock/Dock.tsx?raw"
 
 export const dock = {
-  ...(generateCliCommands('Components/Dock', ['default'])),
+  ...generateCliCommands("Components/Dock", ["default"]),
```
DavidHDev, line 9: Please update this function call to contain all the available variants.

whitedevil-glitch: oh, with the "ts-default" and "ts-tailwind"?

ipuppyyt: > oh, with the "ts-default" and "ts-tailwind"?

yes, just remove ```['default']``` and keep just ```...generateCliCommands("Components/Dock")``` if you have added both `ts-default` and `ts-tailwind`

whitedevil-glitch: yep, done

src/constants/code/Components/dockCode.js
```diff
@@ -1,14 +1,18 @@
...
-import code from '@content/Components/Dock/Dock.jsx?raw';
-import css from '@content/Components/Dock/Dock.css?raw';
+import code from "@content/Components/Dock/Dock.jsx?raw"
+import css from "@content/Components/Dock/Dock.css?raw"
+import tsCode from "@ts-default/Components/Dock/Dock.tsx?raw"
+import tsTailwind from "@ts-tailwind/Components/Dock/Dock.tsx?raw"
 
 export const dock = {
-  ...(generateCliCommands('Components/Dock', ['default'])),
+  ...generateCliCommands("Components/Dock"),
```
DavidHDev, line 9: Still not good, there's no "JS + Tailwind" version of this in the files, so unless you also add that to this PR, this should be:
` ...(generateCliCommands('Components/Dock', ['default', 'ts/default', 'ts/tailwind'])),`

whitedevil-glitch: hmm, i can add the "JS+Tailwind" version as the tailwind styles almost remains the same, sweet i'll commit the changes in a while!

whitedevil-glitch: @DavidHDev  done with the changes mentioned.

PR #73, opened 2025-02-03
chore: Bump jsrepo + add metadata to the manifest

ieedan:
I pretty much just copy pasted the description and tags from github. 

Now when you look at the registry doc it should look like this (minus the localhost):
image


DavidHDev:
That looks good! I can't really think of anything else I would add to the meta info, keep it simple I guess!

DavidHDev (approved):

ieedan:
You should be able to view the updated docs soon putting the final touches on this PR now!

PR #97, opened 2025-02-14
fix: ASCIIText component disposal

Pareder:
- added event listeners cleansing
- added animation frame cancellation
- added scene cleansing
- removed unused code
- removed warning text from "Ascii Text" demo page

Now the component is not laggy and you can easily change the props on demo page and see them in action.


DavidHDev (approved):

DavidHDev:
Awesome, thanks for this!

PR #99, opened 2025-02-14
Fix/squares unwanted rerendering

Saitejavarma30:
added the condition as to when the state has to be changed and accordingly rerendering it


Saitejavarma30:
the fix has been done by @DavidHDev hence closing this pr

PR #100, opened 2025-02-14
Feat/ Squares Multi color on Hover

Saitejavarma30:
- Squares component can take multiple colors instead of 1 and randomly pick the colors when hovered.
- This component used a max(speed, 0.1) which used to make the squares move even when the user gave 0 as the speed. Fixed the issue to allow static squares
- Changed the preview of the squares to use the above feature/fix


DavidHDev:
I really don't like the idea of having multiple colors in the base component, it defeats the purpose of minimalism and makes it feel tacky, (I tried it out locally) so I don't know if I want it to be part of the base component. 

This is something you would bring to your own copy of the code if you wanted to enhance it with something like this for whatever reason, but it just doesn't feel right for this component in particular

Issue #118, opened 2025-02-21
[BUG]: backgrounds/dither rendering issue on some devices


DavidHDev:
### Describe the issue

The problem is that the component does not render well on Windows devices or anything else other than MacOS, I don't know why, but it needs to be investigated and fixed. Opening this issue for visibility.

### Reproduction Link

https://www.reactbits.dev/backgrounds/dither

### Steps to reproduce

Open the page on a Windows device.

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

DavidHDev:
Update: added SOMEWHAT of a fix for this, but it's not super reliable, so I will leave this open in case anyone still encounters problems, please let me know!

DavidHDev:
Hi @nachitodev! Yeah, this is exactly the problem I was talking about, try to play around with the padding of the container. Unfortunately I'm not that good with shaders to figure out what's wrong with this and fix it 100%

mrdbstn:
yup, have the same issue

DavidHDev:
@mrdbstn this whole project is on React 18, versioning has nothing to do with this bug.

mrdbstn:
so, no update on this yet?

DavidHDev:
@mrdbstn if there was an update, I would have closed the issue

PR #159, opened 2025-03-18
Fix Dither

JusGu:
This PR attempts to fix: https://github.com/DavidHDev/react-bits/issues/118

I honestly don't fully understand why this issue only occurred on Mac and not windows but I suspect it has to do with the handling of floating‑point math/ integer conversion. I also unfortunately don't fully understand why this solution works (on a high level, I changed the way the UV pixel was calculated and messed around with the colors until they resembled the original effect) but the you're free to play around with this code.

I was able to reproduce this bug on my Mac with Intel chip which suggests this bug wasn't only on Windows devices. I fixed the bug locally on the same device, I heavily suggest testing this on a Windows device first.

### DIFF
--- a/src/content/Backgrounds/Dither/Dither.jsx
+++ b/src/content/Backgrounds/Dither/Dither.jsx
@@ -114,10 +114,14 @@ const float bayerMatrix8x8[64] = float[64](
 );
 
 vec3 dither(vec2 uv, vec3 color) {
-  int x = int(uv.x * resolution.x) % 8;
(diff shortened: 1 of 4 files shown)


DavidHDev:
Hello @JusGu , really happy to see this PR, thanks a lot! 

I've tested it on Windows and my Mac, and it seems to fix the issues with the shader artifacts and the flickering

DavidHDev (approved):

PR #166, opened 2025-03-25
Update Hyperspeed: adjust renderer size as container size changes

JohnZ03:
Hyperspeed.jsx now listens to the onWindowResize event and adjust the renderer size with respect to the container size accordingly.


DavidHDev:
All the code files should be updated, meaning all four versions of the component.

JohnZ03:
> All the code files should be updated, meaning all four versions of the component.

Ok, I will update the files.

JohnZ03:
@DavidHDev Hi, I have made the change across the versions. Please help double check. Thanks.

PR #169, opened 2025-03-25
Variable proximity calculate on mouse move

PedroMarianoAlmeida:
When I used the `VariableProximity` component, I noticed that my browser started to become slower, so I checked the code and realized that the distance calculation is running all the time and impacts the performance

https://github.com/user-attachments/assets/9a249766-8c8f-4b44-a680-f9b6ee10b3a2

So my fix is to do the same calculation, but only when the mouse moves

https://github.com/user-attachments/assets/aba013d6-2802-4d8b-955e-a955b97dca2a


DavidHDev (approved):
Thanks!

PR #172, opened 2025-03-27
Update package.json with updated dependencies

RJohnPaul:
updated all dependencies


DavidHDev:
Not needed for now, this will be done as part of migrating to React 19.

PR #174, opened 2025-03-30
persist-preferredLanguage

PedroMarianoAlmeida:
Before:

https://github.com/user-attachments/assets/19065704-5958-4902-bba4-b7e2a901b593

After:

https://github.com/user-attachments/assets/9622b897-80dc-433d-8dee-d3812de337a3

Note: The `localStorage.getItem('preferredLanguage') || 'JS';` was already on the code, so I just added the SET  when the user updates the option

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import { Route, BrowserRouter as Router, Routes } from 'react-router-dom'
 import { SearchProvider } from './components/context/SearchContext/SearchContext';
+import { LanguageProvider } from './components/context/LanguageContext/LanguageContext'
 import { useEffect } from 'react';
 import { Toaster } from 'sonner'
 import { forceChakraDarkTheme } from './utils/utils';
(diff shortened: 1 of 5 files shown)


DavidHDev:
Hi @PedroMarianoAlmeida , thanks for the work! The only reason why I didn't do this myself is because I consider that changing the language in the component page should instantly be reflected in the header as well and I didn't allocate time to implement this. Would you be interested in this challenge as part of your PR?

PedroMarianoAlmeida:
Screenshot 2025-03-31 at 3 08 10 PM
 It is that? Yep, I can take care of it on this MR =D

DavidHDev:
Yep that's it :)

PedroMarianoAlmeida:
- I followed the same code style and patterns as in SearchContext. 
- You can review the individual commits to see my step-by-step thought process, but I recommend squashing them into a single commit before merging.
- The result is the screenshot attached 

https://github.com/user-attachments/assets/24904d3b-cb62-414c-8e63-662f27e52b9f

DavidHDev (approved):

DavidHDev:
Thanks! I don't really want to do the same for the styling choice, so I'd say just leave it as is.

PR #187, opened 2025-04-15
updated ShinyText.tsx for tailwind v4

Nishchit-Dev:
Tailwindcss changed the way they declare configuration, now we need to configure using CSS syntax


DavidHDev:
Tailwind v4 migration will all be done in a different future PR.

PR #201, opened 2025-04-29
Added text hover effect with image

Pranaysherkar:
This feature introduces a hover effect for text. When a user hovers over the text:

1. An image (or GIF/video) appears over the text.

2. The image moves along with the mouse while hovering.

3. The image disappears when the mouse leaves the text field.

4. This effect works specifically for the text area, and it can display static images, GIFs, or videos.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1760,6 +1760,7 @@
 			"cpu": [
 				"ppc64"
 			],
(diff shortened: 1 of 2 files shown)


DavidHDev:
The library is not currently accepting new components from contributors.

PR #219, opened 2025-05-24
update CircularText from official page demo - It is not ready but I need a green light

PedroMarianoAlmeida:
# Refactor for Performance Improvement
I refactored the component to improve performance by replacing useState with Framer Motion’s useMotionValue hook. This avoids unnecessary re-renders by preventing state updates on every frame.

If you find this approach effective, I’d be happy to apply the same optimization to the other CircularText components as well (I did the change on the Component that runs on the page, so if you run this branch and navigate to `http://localhost:5173/text-animations/circular-text`, it is possible to see the difference)

Before:

https://github.com/user-attachments/assets/04e1b0a7-623b-48dd-8b6c-4cf7522d67c8

After

https://github.com/user-attachments/assets/bc50aaff-54de-41ed-9c93-eb560d2beccb

### DIFF
--- a/src/content/TextAnimations/CircularText/CircularText.jsx
+++ b/src/content/TextAnimations/CircularText/CircularText.jsx
@@ -1,5 +1,5 @@
-import { useEffect, useState } from "react";
-import { motion, useAnimation } from "framer-motion";
+import { useEffect } from "react";
+import { motion, useAnimation, useMotionValue } from "framer-motion";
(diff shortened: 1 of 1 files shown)


DavidHDev:
Thanks for pointing it out, wasn't able to review on time but I quickly fixed it and and did some other minor tweaks to move things along faster, will close this one

PR #227, opened 2025-06-06
feat: Improve pointerup behaviour

tomasvana10:
Releasing the pointer while interacting with the `Elastic Slider` and `Lanyard` components would cause any elements below your pointer to be clicked. My PR temporarily prevents the element below the mouse from handling the click event (using the `once` attribute), thus introducing a more user-friendly behaviour.


DavidHDev:
This is not needed. Whatever you do with the components in your projects is up to you, they are not made to perfectly work with one another - that's the beauty of copy-paste snippets, they're yours to customize

tomasvana10:
You smell

PR #233, opened 2025-06-07
fix: update useElementWidth to support strict types (fixes #230)

### LINKED ISSUE #230: [BUG]: text-animations/scroll-velocity on TS + Tailwind on ESLint
### Describe the issue

ESLint error 

```
Explain what this problem is and help me fix it: Argument of type 'RefObject<HTMLSpanElement | null>' is not assignable to parameter of type 'RefObject<HTMLElement>'.
  Type 'HTMLSpanElement | null' is not assignable to type 'HTMLElement'.
    Type 'null' is not assignable to type 'HTMLElement'. 
```
The error is because strict types

The fix is to change the *useElementWidth*

After that everything should be fine

### Reproduction Link

_No response_

### Steps to reproduce

1. Install ESLint with strict mode
2. Run `npm run lint` 
3. And there will be an error

to fix that we ened to change `useElementWidth`

to

```tsx
function useElementWidth<T extends HTMLElement>(ref: React.RefObject<T | null>): number {
  const [width, setWidth] = useState(0);

  useLayoutEffect(() => {
    function updateWidth() {
      if (ref.current) {
        setWidth(ref.current.offsetWidth);
      }
    }
    updateWidth();
    window.addEventListener("resize", updateWidth);
    return () => window.removeEventListener("resize", updateWidth);
  }, [ref]);

  return width;
}
```

### Validations

mitran06:
fixed issue #230 by updating the useElementWidth hook to use a generic type, resolving the ESLint strict type error.

### DIFF
--- a/src/ts-default/TextAnimations/ScrollVelocity/ScrollVelocity.tsx
+++ b/src/ts-default/TextAnimations/ScrollVelocity/ScrollVelocity.tsx
@@ -45,7 +45,7 @@ interface ScrollVelocityProps {
   scrollerStyle?: React.CSSProperties;
 }
 
-function useElementWidth(ref: React.RefObject<HTMLElement>): number {
+function useElementWidth<T extends HTMLElement>(ref: React.RefObject<T | null>): number {
   const [width, setWidth] = useState(0);
 
   useLayoutEffect(() => {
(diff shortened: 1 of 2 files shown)


DavidHDev:
Fixes should be applied to both TS files of the component. (CSS & Tailwind files)

PR #276, opened 2025-06-30
react-bits/src/demo/Animation/CrosshairDemo.jsx

chanwoong528:
fixed unexpected several changes on linkText due to width of the text when its near the edge horizontally


chanwoong528:
chrome-capture-2025-06-30
above is the fixed version, 
but in the website if you go closely, it will change like 100 times :)

DavidHDev:
Thanks for this!

Issue #264, opened 2025-06-23
[BUG]: text-animations/circular-text - Tailwind versions are broken


DavidHDev:
### Describe the issue

Using the tailwind components displays the following visual bug:

![Image](https://github.com/user-attachments/assets/1d311247-330d-443e-ad91-e5cab672e21b)

### Reproduction Link

N/A

### Steps to reproduce

Use tailwind/ts-tailwind CircularText component.

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

Mandy98438:
Yo i think i can try some fix here

blacurrant:
@DavidHDev 
Hi David,
I saw that this bug still persisted so I went ahead and fixed it. 
PR: https://github.com/DavidHDev/react-bits/pull/283
Thank you,

PR #283, opened 2025-07-03
Fixed, BUG: text-animations/circular-text - Tailwind versions are broken #264

blacurrant:
Screenshot 2025-07-03 151235

destination: src /tailwind/textAnimations/CircularText

-used tailwindcss to update the Circular Text component
-used the styles from the normal CSS component

PFA, screenshot from the local build.

### DIFF
--- a/src/tailwind/TextAnimations/CircularText/CircularText.jsx
+++ b/src/tailwind/TextAnimations/CircularText/CircularText.jsx
@@ -85,7 +85,7 @@ const CircularText = ({
 
   return (
     <motion.div
-      className={`circular-text ${className}`}
+      className={`m-0 mx-auto rounded-full w-[200px] h-[200px] relative font-bold text-white font-black text-center cursor-pointer origin-center ${className}`}
       style={{ rotate: rotation }}
       initial={{ rotate: 0 }}
       animate={controls}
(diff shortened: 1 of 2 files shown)


src/tailwind/TextAnimations/CircularText/CircularText.jsx
```diff
@@ -85,7 +85,7 @@ const CircularText = ({
 
   return (
     <motion.div
-      className={`circular-text ${className}`}
+      className={`m-0 mx-auto rounded-full w-[200px] h-[200px] relative font-bold text-white font-black text-center cursor-pointer origin-center ${className}`}
```
DavidHDev, line 88: Please remove one:
image

DavidHDev:
Please also update the component under `ts-tailwind` with the same changes.

blacurrant:
Hi, I have updated the components!
Also If you don't mind, why does it need us to change two components for the same?

DavidHDev:
That's because one component uses Typescript, and the other is plain JavaScript. This is because each component has 4 versions that are exported to the CLI registry (jsrepo). Also on the "Code" tab, you can switch in real time between the different versions of the code, allowing developers to use whatever variant they want for a specific component, say for example one of your projects isn't using Typescript, you can copy the plain JS variation instead, etc.

DavidHDev:
Thanks for updating!

Issue #197, opened 2025-04-27
[BUG]: Split Text error


jpsilvestrini:
### Describe the issue

I'm trying to use the split text component in my website, but i had this error:

![Image](https://github.com/user-attachments/assets/6e81a455-150a-4551-9dc0-46ad0a89382b)

### Steps to reproduce

i just put the component, and throw this type i think, error

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

PedroMarianoAlmeida:
On my machine, I don't see the same error, maybe it is a Cursor error, but it is not VS Code (and TypeScript compiler)

Can you try replacing` style={springs[index] as unknown as React.CSSProperties}` by `style={springs[index]}` and check if it works?

If it does not work, try replacing ` ? async (next: (props: { opacity: number; transform: string }) => Promise<void>) => {` by ` ? async (next: (props: any) => Promise<void>) => {` (this was the last change in the code)

Thanks!

jpsilvestrini:
Hello pedro, same error, not solved

DavidHDev:
Wasn't able to reproduce this, I've refactored the component though, you can check if this still happens with the new TS version

DavidHDev:
Will close this for now.

PR #290, opened 2025-07-07
Fix/splittext scroll animation

sudhanshugaikwad:
### Fixes: [#197](https://github.com/DavidHDev/react-bits/issues/197)

This PR addresses the animation bug in the `SplitText` component where animations would not trigger consistently or correctly on scroll.

#### ✅ Changes Made:
- Ensured GSAP and SplitText run only in the browser (using `typeof window`)
- Refactored the `useEffect` to run only after the DOM is ready and text is rendered
- Simplified `ScrollTrigger`'s `start` value to improve reliability
- Improved cleanup logic to prevent memory leaks and animation stacking
- Handled cases where `text` is undefined or updated asynchronously

#### 🧪 Tested:
- Static and dynamic text content
- Scroll behavior on various screen sizes
- Verified that characters animate only once when scrolled into view

#### 💡 Notes:
- This fix ensures robust animation handling for SSR/CSR setups (e.g., Next.js)
- Also improves accessibility by reducing animation jitter or missed triggers

---

Please review and let me know if you need any further changes. 🙌

### DIFF
--- a/src/content/Components/GooeyNav/GooeyNav.jsx
+++ b/src/content/Components/GooeyNav/GooeyNav.jsx
@@ -19,22 +19,13 @@ const GooeyNav = ({
 
   const noise = (n = 1) => n / 2 - Math.random() * n;
 
-  const getXY = (
-    distance,
-    pointIndex,
-    totalPoints
-  ) => {
+  const getXY = (distance, pointIndex, totalPoints) => {
     const angle =
       ((360 + noise(8)) / totalPoints) * pointIndex * (Math.PI / 180);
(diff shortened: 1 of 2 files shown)


DavidHDev:
Any change in a component should be done in all 4 file variants (files under ts-default, ts-tailwind, content, tailwind). This PR will not be merged unless the changes are exactly the same in all files.

DavidHDev:
Added a better refactor for this in https://github.com/DavidHDev/react-bits/commit/fdaaed0bd6092848d81e96e9d10bf4c970d9362b

Issue #292, opened 2025-07-07
[BUG]: Falling Text preview freezing after first render and Rotating Text produced console error


Joyal-George-KJ:
## 1. Falling Text – Preview Freezes on Trigger Change

**Issue**:  
After activating the preview and changing the **Trigger** value in the Falling Text component, the preview stops working. It only activates again if another value (besides Trigger) is changed.

**Steps to Reproduce**:
1. Navigate to the **Falling Text** example.
2. Click **Preview** to activate the animation.
3. Change the **Trigger** value.
4. Attempt to preview again — the animation does **not** run.
5. Change another prop (e.g., Gravity) — now the animation works again.

**Expected Behavior**:  
The animation should re-trigger immediately upon changing the Trigger value.

**Evidence**:  
I’ve recorded a screen capture of the issue. [**Video Link**](https://drive.google.com/file/d/1PQzAvXGP6-7Ip-XSwgnl9dVcOSFjvaCg/view?usp=sharing)

---

## 2. Rotating Text – Invalid HTML Structure

**Issue**:  
In the Rotating Text component, a `<div>` is used as a child of a `<p>` element, which is invalid HTML and throws a error in the console.

**Console Error**:
```log
hook.js:608 In HTML, <div> cannot be a descendant of <p>.
This will cause a hydration error.

  ...
    <TabContent pt={0} value="preview">
      <chakra(TabContent) pt={0} value="preview" css={[...]} ref={null} className="chakra-tab...">
        <Insertion5>
        <TabContent value="preview" className="chakra-tab..." ref={null}>
          <div data-scope="tabs" data-part="content" dir="ltr" id="tabs:«r9»:..." tabIndex={0} ...>
            <PreviewTab>
              <chakra(div) position="relative" className="demo-conta..." minH={400} maxH={400} overflow="hidden">
                <Insertion5>
                <div className="demo-conta..." ref={null}>
                  <div className="rotating-t...">
                    <LayoutGroup>
                      <motion.p className="rotating-t..." layout={true}>
                        <MeasureLayout>
>                       <p className="rotating-text-ptag" style={{}} ref={function}>
                          <motion.span>
                          ...
                            <motion.div className="text-rotate" layout={true} aria-hidden="true">
                              <MeasureLayout>
>                             <div className="text-rotate" aria-hidden="true" style={{}} ref={function}>
```

**Impact**:  
This could cause **hydration mismatches** or layout issues, especially in SSR (Server-Side Rendering) contexts.

[...]

PR #293, opened 2025-07-07
[FIX]: Falling Text preview freezing after first render and Rotating Text produced console error

Joyal-George-KJ:
## 🔧 Description

This PR addresses this issue: #292 

---

### 1. 🧊 Falling Text – Preview Freeze on Trigger Change

**Problem**:  
Changing the **Trigger** value after the initial preview causes the animation to stop working until another unrelated prop is changed.

**Fix**:  
The component state was updated to properly reinitialize the animation whenever the Trigger value changes, ensuring consistent preview behavior.

---

### 2. ⚠️ Rotating Text – Invalid DOM Nesting

**Problem**:  
A `<div>` was used inside a `<p>` tag in the Rotating Text component, causing this console warning:

**Fix**:  
Replaced the nested `<div>` with a `<span>` to ensure valid HTML structure and avoid potential hydration mismatches in SSR environments.

---

## 🧪 Test Plan

- Verified that the **Falling Text** component replays the animation properly on every Trigger change.
- Confirmed that the console no longer shows DOM nesting warnings in the **Rotating Text** component.
- Checked rendering and layout remain consistent after changes.

---

## ✅ Checklist

- [x] Bug fixes are scoped and documented
- [x] Code comments added for clarity
- [x] Console is free of errors and warnings
- [x] Functionality tested locally
- [x] Follows project’s coding standards and contribution guidelines

---

> You can check the code changes, I have ensured everything changed from js/ts default, js/ts tailwind files to ensure that it reflects in the documentation. 1st Bug didn't have to change the all four because it was caused by the demo for Falling Text
> but the second bug I changed all four files.

### DIFF
--- a/src/content/TextAnimations/RotatingText/RotatingText.jsx
+++ b/src/content/TextAnimations/RotatingText/RotatingText.jsx
@@ -168,7 +168,7 @@ const RotatingText = forwardRef((props, ref) => {
     >
       <span className="text-rotate-sr-only">{texts[currentTextIndex]}</span>
       <AnimatePresence mode={animatePresenceMode} initial={animatePresenceInitial}>
-        <motion.div
+        <motion.span
           key={currentTextIndex}
           className={cn(
             splitBy === "lines" ? "text-rotate-lines" : "text-rotate"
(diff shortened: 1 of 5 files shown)


DavidHDev:
Both of these are kinda non-issues, but I'll merge it in. Thanks.

Issue #294, opened 2025-07-07
[BUG]: Gooey Nav - Inconsistent Click and Navigation Behavior in `<li>` with Click Handlers and Styling


Joyal-George-KJ:
### Describe the issue

# In the current component setup of **Gooey Nav**, there is a UX inconsistency involving `<li>` elements that have `onClick` handlers applied, particularly when combined with styling and nested `<a>` tags.

---

### ⚠️ Problem Summary

- When `onClick` is placed **on the `<li>`**, clicking anywhere within the styled `<li>` triggers the **animation/effect**, but **navigation does not occur**.
- When `onClick` is moved **inside the `<a>`** tag:
  - Both the **animation and navigation work**, but The **click target is limited** — users must click **exactly on the anchor text**, not the surrounding area (like padding or borders).
  - This was mentioned in this  Issue #256 but the implementation is false.
- I’ve recorded a screen capture of the issue. [Video Link](https://drive.google.com/file/d/1eJorpapmjHoZbzmUdpcdwhR0molYEOAc/view)

---

### 🖱️ Pointer Device Confusion

- On devices with a pointer (mouse, touchpad, stylus, etc.), the **cursor changes to a pointer (hand)** over the entire `<li>`, giving the impression that the whole area is clickable.
- However, clicking on there (outside the actual `<a>` but inside `<li>`) **does nothing** — no animation, no navigation.
- This results in **poor UX** and can be **confusing to users**, especially when the `<li>` is styled with borders, padding, or background effects.

---

## 💥 Expected Behavior

- The **entire visible `<li>` area** (including padding and borders) should:
  - Trigger the **animation/effect** and  
  - Navigate to the linked page or section when clicked — not just the anchor text itself.

---

> I just wanted to say — I was working on this (related to Issue #256), and I realize I missed the chance to inform @DavidHDev when I should have. I assumed the original PR would be updated and resubmitted, so I held off, but that was my mistake. Apologies for not speaking up earlier.

### Steps to reproduce

Try Clicking on the edge of nav button using mouse (when the cursor changes from default/arrow to pointer/hand) the animation and navigation won't trigger.

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

DavidHDev:
This issue feels AI-generated, can you confirm?

joyalgeorgekj:
Yes I did use AI to generate the description for issue and PR but the bug is real. We discussed it before in another issue #256 

[...]

PR #295, opened 2025-07-07
[FIX}: Gooey Nav - Inconsistent Click and Navigation Behavior in <li> with Click Handlers and Styling

Joyal-George-KJ:
While working on Issue #294, I initially explored enhancing navigation behavior by using the `useNavigate` hook from `react-router-dom` inside the `<li>`'s click handler. 
- It worked well for **route-based navigation**, but:
  - ❌ It did **not support hash-based jumps** (e.g., `#about`, `#home`, `#contact`).
  - Since `useNavigate` doesn't trigger anchor-style scrolling, this approach was dropped before submitting a PR.

---

## 💡 Final Approach

After testing and refining, I went with a cleaner, style-based and DOM-structure approach that solves both navigation and animation issues:

### ✅ 1. Restructured Click Area for Reliable Navigation
- Moved `padding` from the `<li>` to the `<a>` tag.
- Applied `inline-block` to `<a>` so the entire visual area is now **clickable and navigable**, even for hash links.
- This uses native anchor behavior, which properly supports hash scrolling.

### ✅ 2. Fixed Animation Trigger
- Originally, the animation was triggered via an `onClick` on the `<li>`, which no longer worked once interaction was shifted to the `<a>`.
- Moved the `onClick` handler (for animation) to the `<a>`, ensuring both **animation and navigation work** seamlessly.

---

## 🧪 Tested Scenarios

- ✅ Route and Hash Jumps work as expected.
- ✅ Full `<li>` visual area responds to hover, click, and pointer styling.
- ✅ Animation effect triggers consistently.
- ✅ DOM structure is now semantically valid and hydration-safe.

### DIFF
--- a/src/content/Components/GooeyNav/GooeyNav.css
+++ b/src/content/Components/GooeyNav/GooeyNav.css
@@ -33,6 +33,11 @@
   color: white;
 }
 
+.gooey-nav-container nav ul li a {
+  display: inline-block;
+  padding: 0.6em 1em;
+}
+
 .gooey-nav-container nav ul li:focus-within:has(:focus-visible) {
   box-shadow: 0 0 0.5px 1.5px white;
 }
--- a/src/content/Components/GooeyNav/GooeyNav.css
+++ b/src/content/Components/GooeyNav/GooeyNav.css
@@ -25,7 +25,6 @@
 }
 
 .gooey-nav-container nav ul li {
-  padding: 0.6em 1em;
   border-radius: 100vw;
   position: relative;
   cursor: pointer;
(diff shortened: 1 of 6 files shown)


DavidHDev (approved):

DavidHDev:
Looks like you have some conflicts.

DavidHDev:
Also, please don't use prettier in your PRs, it creates a lot of noise.

joyalgeorgekj:
> Looks like you have some conflicts.

@DavidHDev Yeah I didn't see that you pushed the same thing but you didn't solve it. The issue still there! You can check. I can  removed conflict

PR #298, opened 2025-07-09
fix: memory leak in Noise component

mapleleafu:
## Problem
The Noise component has a critical memory leak causing 1GB/minute memory growth. Testing on the official React Bits website showed memory usage increasing from 400MB to 1.4GB within 60 seconds.

memory-usage-screenshot

## Root Cause
The implementation creates a new `CanvasPattern` object every 2 frames via `ctx.createPattern()`. These patterns aren't garbage collected properly, accumulating in memory indefinitely.

## Browser Testing Results
- **Firefox Dev Edition**: ~1GB/minute memory growth (most severe)
- **Chrome/Edge/Safari**: Slower initial growth (10-20s) then rapid acceleration
- All browsers eventually exhibit unbounded memory growth

## Solution
- Fixed canvas size (1024x1024) scaled via CSS instead of using screen resolution
- Direct pixel manipulation with `putImageData()` instead of pattern creation
- Eliminates all pattern objects while preserving the noise effect

## Implementation Details
The `canvasSize` parameter affects both appearance and performance:
- **512px**: Better performance but noticeably larger grain
- **1024px**: Optimal balance - nearly identical to original appearance
- **2048px+**: Closer to original but causes performance issues

After testing various sizes, 1024px provides the best trade-off between visual fidelity and performance.

## Testing
Memory usage remains stable indefinitely with the fix applied:
- Before: 1GB/minute growth
- After: Stable ~50MB usage

## Breaking Changes
None - component API unchanged.

### DIFF
--- a/src/content/Animations/Noise/Noise.jsx
+++ b/src/content/Animations/Noise/Noise.jsx
@@ -12,47 +12,45 @@ const Noise = ({
 
   useEffect(() => {
     const canvas = grainRef.current;
-    const ctx = canvas.getContext('2d');
-    let frame = 0;
+    if (!canvas) return;
 
-    const patternCanvas = document.createElement('canvas');
-    patternCanvas.width = patternSize;
-    patternCanvas.height = patternSize;
-    const patternCtx = patternCanvas.getContext('2d');
-    const patternData = patternCtx.createImageData(patternSize, patternSize);
-    const patternPixelDataLength = patternSize * patternSize * 4;
+    const ctx = canvas.getContext('2d', { alpha: true });
+    if (!ctx) return;
+
+    let frame = 0;
+    let animationId;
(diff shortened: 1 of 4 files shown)


DavidHDev:
Nice, thanks!

mapleleafu:
@DavidHDev Thank you for the quick merge! I really appreciate your work on this library.

While I've done extensive testing on this fix and it resolves the memory leak, I wanted to mention that there might be room for further optimization. The approach trades some visual detail for performance (using a fixed 1024px canvas), and there could potentially be more elegant solutions that maintain the original appearance while preventing the memory leak.

I'd encourage additional testing across different use cases to ensure it meets all performance requirements. Happy to help iterate on this if you find any edge cases or have ideas for improvements.

PR #300, opened 2025-07-09
fix: missing pointer-events-none causing interaction blocking

mapleleafu:
https://github.com/user-attachments/assets/40e2b0fb-f41d-4246-9cb3-f6fbdbd555fe

- Added pointer-events: none in all versions
- Prevents Noise overlay from blocking mouse interactions with underlying components
- Ensures components like TiltedCard remain interactive when Noise is overlaid


DavidHDev:
Pls resolve your conflicts

PR #308, opened 2025-07-14
feat: Added decimal numbers support

AhmedRadwan4:
Setting the "to" argument to a decimal value will add a float point while counting instead of rounding to the nearest number


DavidHDev:
If you modify a component, all 4 versions should be updated with exactly the same code.

AhmedRadwan4:
> If you modify a component, all 4 versions should be updated with exactly the same code.

All 4 versions have been successfully edited to have the same code !

PR #316, opened 2025-07-17
Update Orb.jsx

Aadvait-Hirde:
I optimized the fragment shader with quality controls, added quality based pixel density, throttled mouseevents for better performance, and added an intersection observer for better performance when off-screen.

### DIFF
--- a/src/content/Backgrounds/Orb/Orb.jsx
+++ b/src/content/Backgrounds/Orb/Orb.jsx
@@ -1,14 +1,24 @@
 import { useEffect, useRef } from "react";
 import { Renderer, Program, Mesh, Triangle, Vec3 } from "ogl";
-import "./Orb.css";
+
+interface OrbProps {
+  hue?: number;
+  hoverIntensity?: number;
+  rotateOnHover?: boolean;
+  forceHoverState?: boolean;
+  scale?: number;
+  quality?: 'low' | 'medium' | 'high';
+}
 
 export default function Orb({
   hue = 0,
   hoverIntensity = 0.2,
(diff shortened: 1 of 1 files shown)


DavidHDev:
Please take time to understand the structure of the project before opening PRs. It seems that your code changes are AI-generated, because I don't understand why you would willingly add TypeScript code in a JSX file, it wouldn't even compile in the live server.

Keep TS in TSX files and JS in JSX files. The same goes for CSS and Tailwind. Files under /content should ONLY use CSS, while files under /tailwind do not have an external CSS file.

Besides that, if you want to help improve this, all 4 versions of this component (/content, /tailwind, /ts-default, /ts-tailwind) should be modified to contain the exact same code changes. This is also noted in CONTRIBUTING.md

Other things to watch out for:
- Don't add comments
- Ensure that the component behaves and looks exactly as it did before your changes
- Post screenshots

Aadvait-Hirde:
Hey David, I'd like to apologize. I'm pretty new to contributing to open-source so I should've taken the time to understand the structure properly. I stupidly forgot that this was a jsx file and not a tsx file since I cloned it off of my local repo which I was testing on, but you're completely right, my bad again!

DavidHDev:
Not a problem, you're welcome to contribute, I just don't understand how you were able to test these changes since even defining an interface in a JSX file crashes the whole compilation.

DavidHDev:
Oh, you tested it in a local project. All changes should be tested on the actual cloned React Bits project, it's the way to go if we want to ensure the best quality is delivered

PR #386, opened 2025-08-13
fix: Enhance CircularGallery: enable mipmaps for textures for smooth edge 

minwoogwak99:
Enabled antialiasing and mipmaps for better texture at edge of each images with smooth.

### AS-IS
Screenshot 2025-08-12 at 11 50 08 PM

### TO-BE
Screenshot 2025-08-12 at 11 50 21 PM


DavidHDev:
Awesome, thanks!

PR #402, opened 2025-08-18
TextType: add useMemo and useCallback to improve performance

WarriorAchilles:
The `getRandomSpeed` function was being created on every render, and the `textArray` variable was being recalculated on every render. This wraps `textArray` in a `useMemo` hook and wraps `getRandomSpeed` in a `useCallback` hook to improve the performance of the component.


DavidHDev:
Cool, thanks!

PR #404, opened 2025-08-19
fix(rotatingtext): replace AnimationControls with LegacyAnimationCont…

nmsn:
Replace deprecated AnimationControls with LegacyAnimationControlsto fix TypeScript compilation errors.


DavidHDev:
No need to add legacy imports in the files. The correct fix is to just remove `AnimationControls`, I've already pushed it.

PR #407, opened 2025-08-20
Fix: Add null checks for rendererRef and canvasRef to prevent server crash

### LINKED ISSUE #406: [BUG]: Server crash due to renderRef in PlasmaWaveV2 component
### Describe the issue

The server goes down when rendering the PlasmaWaveV2 component.
The issue seems to be caused by the renderRef implementation inside the component.

### Reproduction Link

_No response_

### Steps to reproduce

	1.	Start the development server (npm run dev or yarn dev).
	2.	Import the PlasmaWaveV2 component into any page or test component.
	3.	Render <PlasmaWaveV2 /> inside your application.
	4.	Observe the server logs after the component mounts.

Result: 
- The server crashes immediately due to the renderRef usage in PlasmaWaveV2.

### Validations

L1m3Kun:
This PR fixes server crashes caused by missing ref objects in two components:
	•	PlasmaWaveV2: Added null check for rendererRef.
	•	LetterGlitch: Added null check for canvasRef.

Previously, when either rendererRef or canvasRef was null, the server would crash unexpectedly. With these changes, both components now safely exit early if the ref is not available.

Closes #406

### DIFF
--- a/src/components/landing/PlasmaWave/PlasmaWaveV2.jsx
+++ b/src/components/landing/PlasmaWave/PlasmaWaveV2.jsx
@@ -160,7 +160,9 @@ export default function PlasmaWaveV2({
     if (isMobile) {
       return;
     }
-
+    if (!rendererRef.current) {
+      return;
+    }
     const renderer = new Renderer({
       alpha: true,
       dpr: Math.min(window.devicePixelRatio, 1),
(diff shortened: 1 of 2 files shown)


DavidHDev:
The PlasmaWave is not part of the public component list, but okay, I'll merge this.

L1m3Kun:
Thank you for merging. 
While using reactbits.dev, I encountered a null ref issue that caused some pages to not work as expected. I truly appreciate this project, and I believe this fix will help improve its stability. 🙏

PR #426, opened 2025-08-26
Feat/upgrade splittext component

Toosoo:
- adding useGSAP hook from gsap for better scoping and cleanup also works better with Nextjs
- added tag option for better SEO and performance 
- using onSplit for better performance and responsiveness
- listening for fonts loading which recommended from gsap to prevent CLS and sizes issues
- fix words break to next line when using only chars to split by adding smartWrap option from gsap 
- fix lines wrapping while resizing by using autoSplit from gsap 
- switch to fromTo() instead of set() to()  
- remove extra code like toggleActions , immediateRender , force3d , ...... 
- adding classes for chars , words and lines

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
 			"dependencies": {
 				"@chakra-ui/react": "^3.20.0",
 				"@emotion/react": "^11.14.0",
+				"@gsap/react": "^2.1.2",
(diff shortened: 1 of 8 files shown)


DavidHDev (approved):
Nice, thank you!

PR #433, opened 2025-08-28
fix: suppress unused parameter warning in hex shorthand

EnderRomantice:
Renames `m` to `_m` in replace callback to indicate intentional non-use and avoid TS `noUnusedParameters` warning.


DavidHDev:
There's a TSX file under /ts-default where this change needs to be replicated, please do that and we can merge it.

EnderRomantice:
@DavidHDev Thanks for the feedback! I've applied the same fix to the TSX file in the /ts-default directory and pushed the update. The change is now included in this PR.

Please let me know if there's anything else that needs adjustment. Appreciate your review!

DavidHDev:
Thanks!

PR #435, opened 2025-08-28
fix: Target cursor issue (from Vue Bits)

xzadudu179:
Fixed the same problem I solved in [Vue Bits #84](https://github.com/DavidHDev/vue-bits/issues/84)


DavidHDev:
This time it should be in all 4 variants of the component :)

xzadudu179:
it should be fine now

DavidHDev (approved):

PR #436, opened 2025-08-29
Fix CardSwap animation by animating on z axis

BartaBlzs:
When the lower section of the cards are visible the animation feels snappy. By animating on the z and x axis it fixes that.


DavidHDev:
All 4 variants of the component should be updated with the same code.

PR #494, opened 2025-09-14
fix: resolve Dither background scroll flickering issue

Shubhamkanskar:
- Add scroll-based animation pausing to prevent flickering during scroll
- Implement autoPauseOnScroll, scrollPauseThreshold, and resumeOnScrollUp props
- Freeze animation time during scroll events to maintain visual consistency
- Add scroll event listener with passive option for better performance
- Apply fix to all 4 Dither component variants (JSX, TS+CSS, TS+TW, JS+TW)
- Resolves performance issue where intensive shader calculations conflicted with browser scroll rendering

Fixes: Background flickering during scroll that made the component unusable

### DIFF
--- a/src/content/Backgrounds/Dither/Dither.jsx
+++ b/src/content/Backgrounds/Dither/Dither.jsx
@@ -174,21 +174,11 @@ function DitheredWaves({
   pixelSize,
   disableAnimation,
   enableMouseInteraction,
-  mouseRadius,
-  autoPauseOnScroll = true,
-  scrollPauseThreshold = null,
-  resumeOnScrollUp = true
+  mouseRadius
 }) {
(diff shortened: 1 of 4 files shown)


DavidHDev:
Are you sure you checked this? While testing it locally, it did not seem to make any visible difference or do anything to fix the flickering issue. 

Plus, a pause on scroll functionality has nothing to do with this, and does not fix the actual problem we have here (and it doesn't even work)

DavidHDev (changes requested):
Changes do not fix the issue. Pause on scroll is not the solution.

Shubhamkanskar:
> Changes do not fix the issue. Pause on scroll is not the solution.

ok I'll try different approach come up with new fix

Issue #535, opened 2025-09-30
[BUG]: the fuzzy-text at text-animations


ashykng:
### Describe the issue

I found 2 problems or bug in this component:

u can not separate the text into two lines
I tried with \n, ```<br />``` and also passing the separate lists too

the text does not showing completely and also the canvas is not centered and it hasn't the full-width and full-height

### Steps to reproduce

using the "code" section with type of typescript + tailwind

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

ashykng:
these problems and bugs have fixed in pull request #537

now with these changes u can separate the text with ```<br />``` tag

and also the canvas is centered and also it gets the width and height from the parent

DavidHDev:
Not a bug.

PR #537, opened 2025-09-30
fix for bug #535

ashykng:


DavidHDev:
Not a bug. If you want multiple lines, use multiple instances of the component. This code is unreliable.

PR #564, opened 2025-10-11
Fix: color changing logic for LightRays Background

SergeDeGuzman:
The hexToRgb constant expects exactly 6 hex digits and when passed by an 8-digit string, it does not match and the function returns [1,1,1] which is white. Also, some WebGL/ogl code prefers typed arrays for uniform vectors, so it's safer to set uniform values as Float32Array. I replaced your hexToRgb with this function that supports #RGB, #RRGGBB, #RRGGBBAA. I find this minor bug when I'm making my portfolio website and I like to thank you for having this open source codes.

### DIFF
--- a/src/content/Backgrounds/LightRays/LightRays.jsx
+++ b/src/content/Backgrounds/LightRays/LightRays.jsx
@@ -4,9 +4,29 @@ import './LightRays.css';
 
 const DEFAULT_COLOR = '#ffffff';
 
-const hexToRgb = hex => {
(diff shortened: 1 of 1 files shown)


DavidHDev:
Any component code change should be added to all 4 variants, meaning 4 files changed in this PR.

Please read CONTRIBUTING.md

SergeDeGuzman:
I only changed a constant in content/Backgrounds/LightRays/LightRays.jsx

DavidHDev:
@SergeDeGuzman this component lives under 3 other files as well, they all need to be kept in sync.

PR #588, opened 2025-10-19
Resolving component import errors

### LINKED ISSUE #587: [BUG]: components/pill-nav
### Describe the issue

On importing the component, it says that the onMobileMenuClick is not used while importing the component, which is a required property!

Image

## Terminal error:

Image

### Reproduction Link

_No response_

### Steps to reproduce

Please let me know if I can raise a PR!

### Validations

ashutoshdebug:
I have changed the import component of the PillNav. Now it contains all the properties which are required at the time of importing and using!

It fixes #587

Code-snippet:
image

File path is: \react-bits\src\constants\code\Components\pillNavCode.js

### DIFF
--- a/src/constants/code/Components/pillNavCode.js
+++ b/src/constants/code/Components/pillNavCode.js
@@ -25,6 +25,7 @@ import logo from '/path/to/logo.svg';
   pillColor="#ffffff"
   hoveredPillTextColor="#ffffff"
   pillTextColor="#000000"
+  onMobileMenuClick
 />`,
   code,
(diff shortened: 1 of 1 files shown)


DavidHDev:
This doesn't fix anything. What you see in pillNavCode.js is only the snippet displayed in the demo page, not the actual component code.

ashutoshdebug:
I was saying that if someone uses this displayed snippet in their code, then it might throw an error for the onMobileClickMenu property is not present!

PR #597, opened 2025-10-21
Fix(TextType): Inherit Text Color When Not Provided

columk1:
## Changes

* Changed `getCurrentTextColor()` to return `undefined` instead of `#ffffff` when no `textColors` are provided

## Why

Previously, the component defaulted to white when `textColors` was empty, preventing color inheritance from parent elements. This gives the component the expected drop-in behaviour.

### DIFF
--- a/src/content/TextAnimations/TextType/TextType.jsx
+++ b/src/content/TextAnimations/TextType/TextType.jsx
@@ -42,7 +42,7 @@ const TextType = ({
   }, [variableSpeed, typingSpeed]);
 
   const getCurrentTextColor = () => {
-    if (textColors.length === 0) return undefined;
+    if (textColors.length === 0) return;
     return textColors[currentTextIndex % textColors.length];
   };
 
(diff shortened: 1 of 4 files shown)


src/content/TextAnimations/TextType/TextType.jsx
```diff
@@ -42,7 +42,7 @@ const TextType = ({
   }, [variableSpeed, typingSpeed]);
 
   const getCurrentTextColor = () => {
-    if (textColors.length === 0) return '#ffffff';
+    if (textColors.length === 0) return undefined;
```
DavidHDev: No need to return a value, just call return, it's the same as undefined.

Also, this fix is not complete. Where we use `getCurrentTextColor` in the component below, it should look like this to specify the "inherit" behavior:

```js
    <span className="text-type__content" style={{ color: getCurrentTextColor() || 'inherit' }}>
```

Issue #642, opened 2025-10-31
[BUG]: Target Cursor does not snap properly on mobile/touch devices


Velltrix:
### Describe the issue

When using the target cursor component on a mobile device and clicking on a button, the cursor does not snap properly to the buttons.

![Image](https://github.com/user-attachments/assets/e185c876-b9fc-457c-9e0e-f18cd50d3205)
<img width="939" height="730" alt="Image" src="https://github.com/user-attachments/assets/7d833977-c20b-48c6-9d1e-877a10518c0f" />

### Steps to reproduce

Open https://www.reactbits.dev/animations/target-cursor on a mobile device and click on any of the buttons.

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

PR #643, opened 2025-10-31
FIX: Fixed mobile bug by adding proper handling

Velltrix:
This PR fixes issue #642

Problem: on touch/mobile, clicking a button caused the TargetCursor to snap to an incorrect place.

react-bits-bug-report_1
react-bits-bug-report_2

Root cause: mobile devices don’t reliably emit mousemove events. 

What I changed (applied to all 4 variants):
- Added lastPosRef initialized to window center.
- Updated moveCursor to store lastPosRef.current = { x, y } whenever the cursor moves.
- Added a pointermove handler registered on window that calls moveCursor(e.clientX, e.clientY) so touches update the stored position.
- Added pointerdown/pointerup equivalents for mousedown/mouseup so touch press/release animations run.
- When entering a target, call updateCorners(lastPosRef.current.x, lastPosRef.current.y) if no mouse event is present (fallback to stored pos).
- Remove pointer listeners (pointermove/pointerdown/pointerup) in the effect cleanup.

Why this fixes the bug:
Touch devices fire pointer events but not mousemove, so the code now listens to pointermove to keep lastPosRef updated for touch moves and taps. When a target is entered without a mouse event (for example on tap), updateCorners is given the stored lastPosRef coordinates instead of undefined or a stale value. Pointerdown/pointerup handlers were added so the click-style press animation runs on touch devices as well.

Outcome:
react-bist-bug-report-fix_2
react-bist-bug-report-fix_1

### DIFF
--- a/src/content/Animations/TargetCursor/TargetCursor.jsx
+++ b/src/content/Animations/TargetCursor/TargetCursor.jsx
@@ -7,6 +7,7 @@ const TargetCursor = ({ targetSelector = '.cursor-target', spinDuration = 2, hid
   const cornersRef = useRef(null);
   const spinTl = useRef(null);
   const dotRef = useRef(null);
+  const lastPosRef = useRef({ x: window.innerWidth / 2, y: window.innerHeight / 2 });
   const constants = useMemo(
     () => ({
       borderWidth: 3,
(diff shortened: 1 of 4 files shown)


DavidHDev:
I'm not sure your changes actually fix the problem, since I'm still able to reproduce the cursor not locking perfectly on tap using your version of the component:

Example 1:

image

<hr />

Example 2: 

image

Velltrix:
Okay, I couldn’t reproduce the issue with the cursor not locking after the changes, but I’ll keep trying to reproduce it and look for another fix.

DavidHDev:
I think the real (and simplest) solution here is to disable this effect on mobile devices, since it makes no sense for a cursor-following component to be used on a mobile device. I pushed this as a default behavior.

PR #648, opened 2025-11-01
Fix CountUp Initial Formatting

nathansbird:
Currently, the formatting on the CountUp component does not work if the animation never runs. This means providing two equivalent to/from values > 1,000 results in a lack of formatting.

For example:
To: 15,000
From: 15,000
Displays: 15000

As soon as you change 'from', the spring values are formatted properly.

I fixed this by extracting the formatting logic and applying it to the initial text display setter.

This is great for using the CountUp more as a passively animated component. The text only animates when there's a change in values and doesn't overwhelm the viewer if used in lots of places.


DavidHDev:
Thanks!

Issue #665, opened 2025-11-05
[BUG]: Variable declared but not used


Darshan98Solanki:
### Describe the issue

hello @mapleleafu @Jakjm @agustinusnathaniel @nathansbird @saadh393 

In GridScan Component the variable res created but not used.
please assign me this issue to remove just one line -_-'

<img width="1440" height="157" alt="Image" src="https://github.com/user-attachments/assets/066e4433-5b5c-4390-aaf2-dd41259b2d17" />

### Steps to reproduce

NA

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

DavidHDev:
Please don't tag people randomly in the issue. 

Also, if you want to fix something, just do it, you don't need to be assigned for one line, it's just a formality

Darshan98Solanki:
@DavidHDev sorry for that I'll raising pr in few mins.

PR #666, opened 2025-11-05
fix: remove an extra var not needed just for formality

Darshan98Solanki:
fix: #665 

hey @DavidHDev 

remove that extra variable from ts-default and ts-tailwind for grid scan

image

sorry for tagging multiple people on bug.


DavidHDev:
Haha, I was saying that the "assignment" part on the issue is a formality, not the code. Thanks!

PR #668, opened 2025-11-06
feat: Improve, Refactor Target Cursor

son426:
This PR includes:

1. fix: Remove duplicate mousemove event listener
   - Minor bug fix for accidentally duplicated event listener

2. feat: Refactor animation logic and add customization options
   - Switched to gsap.ticker for better performance and smoother animations
   - Unified animation control using activeStrengthRef-based interpolation
   - Simplified parallax effect using animation duration control instead of position offsets
   - Added `hoverDuration` and `parallaxOn` props for customization

---

**Note:**

I've structured this refactor with performance and clarity in mind. Please let me know if you have a different direction or any feedback on the architecture.

On a separate note, I saw that the component no longer supports mobile. Could you clarify our long-term plan for this? If mobile support is still a goal, I'm happy to tackle the positioning bug in a follow-up PR.

Thanks!

### DIFF
--- a/src/constants/code/Animations/targetCursorCode.js
+++ b/src/constants/code/Animations/targetCursorCode.js
@@ -14,6 +14,7 @@ export default function App() {
       <TargetCursor 
         spinDuration={2}
         hideDefaultCursor={true}
+        parallaxOn={true}
       />
       
       <h1>Hover over the elements below</h1>
--- a/src/content/Animations/TargetCursor/TargetCursor.jsx
+++ b/src/content/Animations/TargetCursor/TargetCursor.jsx
@@ -2,28 +2,36 @@ import { useEffect, useRef, useCallback, useMemo } from 'react';
 import { gsap } from 'gsap';
 import './TargetCursor.css';
 
-const TargetCursor = ({ targetSelector = '.cursor-target', spinDuration = 2, hideDefaultCursor = true }) => {
+const TargetCursor = ({
+  targetSelector = '.cursor-target',
+  spinDuration = 2,
+  hideDefaultCursor = true,
+  hoverDuration = 0.2,
+  parallaxOn = true
+}) => {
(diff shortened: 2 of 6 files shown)


DavidHDev:
It makes no sense to include support for this on mobile, it being a cursor component, as a user, I would find it annoying to have a box floating around on my screen and for it to only move around when I tap on things I don't even know are targets, to see the corners snap.

So the only possible variant where this would work on mobile is if we only include the corner snapping animation without the cursor itself, but that seems pretty pointless.

Other than that, thanks for the updates. LGTM.

son426:
Thanks. Quick question regarding future PRs: Are you generally okay with me tweaking default props in demos to enhance the initial user experience (e.g., for this PR, I considered defaulting parallaxOn to false)? Let me know if that's a welcome type of change.

DavidHDev:
@son426 anything you want to modify, you can ask in the corresponding issue beforehand, and I will either say yes or no

son426:
Thanks for clarifying the workflow. So just to be sure, do you prefer an issue first not just for subjective changes like props, but for all types of PRs, including performance improvements and smaller bug fixes like in this one? I just want to make sure I'm not flooding the repository with issues.

On a related note, do you prefer separate commits for distinct changes within a single PR (e.g., a fix in one commit, a refactor in another)?

DavidHDev:
Anything that you want to do should be discussed, otherwise you risk having your PRs closed. I'm very conservative with that goes into the codebase, so don't be surprised if I say no to your issues lol

DavidHDev:
Also, I don't care about commits, do it however you like

son426:
Alright, crystal clear. I'll be sure to open an issue to discuss any changes beforehand, as you suggested.

I really appreciate the direct feedback and welcome any future pointers on my workflow. See you around! lol

PR #686, opened 2025-11-10
WIP: jsrepo v3 (alternative)

ieedan:
Alternative to #684 

Go with the big ol' file approach for this one. This one requires no changes to any CLI commands (other than jsrepo).

jsrepo and shadcn can actually share the same registry now since none of the registry items depend on eachother and there are no imports to resolve.

- Moved the `registry.json` config -> `jsrepo.config.ts`.
- Removed `jsrepo-build-config`

A few notes:
- Right now the shadcn adapter is set to `format: true` this is nice for readability but it would shrink the registry a decent amount if we turn it off

TODO:
- [ ] Update jsrepo commands
The new commands would look like:
```sh
jsrepo add https://reactbits.dev/r/SplitText-TS-TW
```


DavidHDev:
Seems much better, but may I ask why the content of the registry items has changed?

ieedan:
> Seems much better, but may I ask why the content of the registry items has changed?

Just formatting and key ordering differences should function exactly the same as before

PR #693, opened 2025-11-12
refactor(LogoLoopDemo): Simplify and align demo with component API

son426:
Follow-up cleanup for feature enhancement #682 

- Remove hardcoded 'renderItem' example and 'useCustomRender' switch
  : The predefined function wasn't visible to users, and documentation is clear enough for this optional feature.
- Remove redundant 'LogoLoop' repetition from usage code
- Clean up minor details

### DIFF
--- a/src/constants/code/Animations/logoLoopCode.js
+++ b/src/constants/code/Animations/logoLoopCode.js
@@ -25,7 +25,6 @@ const imageLogos = [
 function App() {
   return (
     <div style={{ height: '200px', position: 'relative', overflow: 'hidden'}}>
-      {/* Basic horizontal loop */}
       <LogoLoop
         logos={techLogos}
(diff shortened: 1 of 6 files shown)


DavidHDev:
I don't see the point of this PR, you're just reordering some stuff and removing useful information from the demo. Closing.

son426:
Thank you for the quick review. Here's a clarification of the changes:

1. The code for the purple-bordered renderItem style was hidden inside the demo file, which I felt could be confusing for users since they couldn't see how it worked.
2. The "Usage" code block contains a duplicated LogoLoop component.

If it's better to show many patterns in the 'Usage' section, I propose adding the renderItem pattern as a commented-out example to the code block.

If you approve of this direction, I will create a new PR.

PR #702, opened 2025-11-14
Refine logoloop demo

son426:
This is a couple of updates for the demo page.

1. Moves the renderItem implementation from being hidden in the demo to the explicit code usage block.
2. Adds a preview switch to control the fadeOutColor.


src/demo/Animations/LogoLoopDemo.jsx
```diff
@@ -262,6 +242,20 @@ const LogoLoopDemo = () => {
             }}
           />
 
+          {fadeOut && (
```
DavidHDev: Showing this in the demo itself isnt really beneficial, since the fade out color only looks good if it matches the background, which is always the same on React Bits. We should remove this from the demo

PR #705, opened 2025-11-14
Update/change conditional, add a condition for when removing the word.

matheussantosspbr:
By using this method, you force the user to submit an array containing two strings for the loop to function correctly (if the loop is active).

```
else if (textArray.length > 1) {
    timeout = setTimeout(() => {
        setIsDeleting(true);
    }, pauseDuration);
}
```

Considering there are two strings in the array, if the loop is deactivated, the component will render both strings and then delete them.

With the change below, I allowed either one string or one string to be passed within the array.

I also fixed the issue so that when the user puts two strings in the array and the loop is deactivated, it will render both strings and not delete the last one, leaving the last string visible on the screen.

```
else if (textArray.length >= 1) {
    if (!loop && currentTextIndex === textArray.length - 1) return;
    timeout = setTimeout(() => {
        setIsDeleting(true);
    }, pauseDuration);
}
```

If you have any questions, feel free to ask. 🙃

### DIFF
--- a/src/tailwind/TextAnimations/TextType/TextType.jsx
+++ b/src/tailwind/TextAnimations/TextType/TextType.jsx
@@ -113,7 +113,8 @@ const TextType = ({
             },
             variableSpeed ? getRandomSpeed() : typingSpeed
           );
-        } else if (textArray.length > 1) {
+        } else if (textArray.length >= 1) {
+          if (!loop && currentTextIndex === textArray.length - 1) return;
           timeout = setTimeout(() => {
             setIsDeleting(true);
           }, pauseDuration);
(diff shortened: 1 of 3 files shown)


DavidHDev:
I don't have questions, but you should update all variants of this component to get this merged, they need to be in sync.

matheussantosspbr:
I only used variables that had already been created and were being used in the useEffect to perform the validation. This change won't directly affect the user. I'm even using it on my website myself.

DavidHDev:
@matheussantosspbr I don't think I've made myself clear. Every component in React Bits has 4 variants. If you update one, you'll have to do the same for the other 3:

image

matheussantosspbr:
I understand, I just implemented that in all 4 variations. I ended up creating 4 commits because I did everything here on the GitHub platform. But it's already done.

PR #723, opened 2025-11-21
Refactor AnimatedContent: custom scroller support, safer cleanup & exit animations

freakdaniel:
Since I use AnimatedContent in my projects, I have been modifying the component to suit my needs as I use it. For this reason, I decided to make this pull request, as I believe that enough has been fixed and added that beginners would really need

### KEY CHANGES
1. Safe Cleanup
**Before:** The component used `ScrollTrigger.getAll().forEach(t => t.kill())` on unmount. This killed ALL scroll triggers on the page, breaking animations in other components
**After:** Now correctly stores references to the specific timeline and trigger instance, killing only its own processes

2. Custom Scroller Support
**Added logic to detect specific scroll containers** (via container prop or snap-main-container id)
Essential for layouts using overflow: scroll, CSS Scroll Snap, or third-party scroll libraries where the window is not the scroller

3. Exit Animations
**Introduced disappearAfter, disappearDuration, and disappearEase props.** Elements can now animate out automatically after the entry animation completes

4. Technical
**Switched** from `gsap.to` to `gsap.timeline` for better control over animation sequencing (Entry -> Delay -> Exit).

### Testing
1. Verify that animations still work in the main window flow
2. Test inside a nested scroll container (pass the container ref)
3. Navigate away from the page and back to ensure other animations are not broken (cleanup check)
4. Test the disappearAfter={2} prop to see the exit animation

### DIFF
--- a/src/content/Animations/AnimatedContent/AnimatedContent.jsx
+++ b/src/content/Animations/AnimatedContent/AnimatedContent.jsx
@@ -21,7 +21,9 @@ const AnimatedContent = ({
   disappearDuration = 0.5,
   disappearEase = 'power3.in',
   onComplete,
-  onDisappearanceComplete
+  onDisappearanceComplete,
+  className = '',
+  ...props
 }) => {
   const ref = useRef(null);
 
@@ -93,7 +95,16 @@ const AnimatedContent = ({
     disappearEase, onComplete, onDisappearanceComplete
   ]);
 
-  return <div ref={ref} style={{ visibility: 'hidden' }}>{children}</div>;
+  return (
+    <div 
+        ref={ref} 
+        className={className}
+        style={{ visibility: 'hidden' }} 
+        {...props}
+    >
+        {children}
+    </div>
+  );
 };
 
 export default AnimatedContent;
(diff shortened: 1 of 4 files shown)


DavidHDev:
All 4 variants need to be kept in sync, please also modify the other 3 if you want to get this merged 😁

freakdaniel:
> All 4 variants need to be kept in sync, please also modify the other 3 if you want to get this merged 😁

It seems that I don't fully understand what synchronization is being discussed, and what options are mentioned in the message. Please explain to me

DavidHDev:
Each React Bits component comes in 4 variants:
- JS with CSS (your changes have been made here) - /content
- JS with Tailwind - /tailwind
- TS with CSS - /ts-default
- TS with Tailwind - /ts-tailwind

When making a change in one of the components, all 4 variants should be kept in sync with the new modified code.

freakdaniel:
> Each React Bits component comes in 4 variants:
> 
>     * JS with CSS (your changes have been made here) - /content
> 
>     * JS with Tailwind - /tailwind
> 
>     * TS with CSS - /ts-default
> 
>     * TS with Tailwind - /ts-tailwind
> 
> 
> When making a change in one of the components, all 4 variants should be kept in sync with the new modified code.

Done

PR #731, opened 2025-11-24
Refactor FadeContent: GSAP migration, seconds units (AnimatedContent-like) and custom scroller support

freakdaniel:
I have modified FadeContent to match the architecture and reliability of the main animation component (AnimatedContent). This PR contains changes for all 4 component variants (JS + CSS/JS + TW, TS + CSS/TS + TW)

### So, again, KEY CHANGES

1. Architecture
**Before:** Used IntersectionObserver mixed with CSS Transitions and setTimeout. duration was in milliseconds
**After:** Refactored to use GSAP Timeline + ScrollTrigger. Duration and delay are now in seconds (e.g., 1 instead of 1000) to match GSAP standards and AnimatedContent

2. Safe Cleanup
Implements st.kill() and gsap.killTweensOf(), ensuring no memory leaks or zombie animations when the component unmounts

3. Custom Scroller Support
**Added logic to detect specific scroll containers** (via container prop or #snap-main-container id)
**Essential for layouts using overflow:** scroll, CSS Scroll Snap, or third-party scroll libraries where the window is not the scroller

4. Exit Animations
**Introduced disappearAfter, disappearDuration, and disappearEase props.** Elements can now fade out (and re-apply blur) automatically after the entry animation completes

5. Technical
**Added visibility:** hidden (or invisible class) to prevent the unstyled content from flashing before GSAP initializes

### Testing

1. Verify that animations work inside nested scroll containers
2. Navigate away from the page and back to ensure no errors occur in the console (cleanup check)
3. Test disappearAfter={2} to see the element fade out and blur again

### DIFF
--- a/src/constants/Information.js
+++ b/src/constants/Information.js
@@ -52,7 +52,7 @@ export const componentMetadata = {
   },
   'Animations/FadeContent': {
     videoUrl: '/assets/video/fadecontent.webm',
-    description: 'Simple directional fade / slide entrance wrapper with threshold-based activation.',
+    description: 'Simple directional fade / slide entrance / exit wrapper with threshold-based activation.',
     category: 'Animations',
     name: 'FadeContent',
     docsUrl: 'https://reactbits.dev/animations/fade-content',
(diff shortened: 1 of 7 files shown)


freakdaniel:
Forgot to update landing in the previous commit :p

I'll update Fade Content too in a while

freakdaniel:
@DavidHDev done with this one

DavidHDev:
Looks good!

PR #746, opened 2025-11-28
refactor(SplashCursor): resolve all ESLint and TypeScript warnings

1serkan:
- replace let with const where variables are not reassigned
- remove all remaining 'any' types and add proper custom types
- fix "pointers" and "uniforms" declarations to follow prefer-const


DavidHDev:
the same changes should be added under the /ts-tailwind variant

PR #757, opened 2025-12-03
perf(Shuffle): Optimize component with memoization

EnderRomantice:
Cache scroll trigger start value:

The calculation of the scroll trigger start position (involving regular expressions and string concatenation) is now cached. It is only recalculated when the threshold or root margin property changes.

Cache render-time values:

Values ​​created on each render are now cached to ensure reference stability ->

userHasFont boolean check.

fallbackFont style object.

The final commonStyle object passed to the element.

classes string.

Fix useGSAP dependency:

When a parent component provides a new callback function, the effect might not re-run correctly, triggering an outdated closure issue.

Therefore, I have added the callback function to the useGSAP dependency array.


EnderRomantice:
Thanks for your merge :P

David ***@***.***> 于2025年12月5日周五 00:06写道：

PR #763, opened 2025-12-05
perf: Optimize animation loop by reducing unnecessary DOM writes and type safe for TextPressure

EnderRomantice:
The debounce logic for the resize event was ineffective due to an unstable function reference, and the dynamic <style> tag was being recreated on every render, causing unnecessary performance overhead.

So I used useCallback: Stabilized the reference of the setSize function to ensure debounce works correctly for the resize event.
and used useMemo: Cached the <style> tag to prevent it from being re-inserted into the DOM on every render.
Other Optimizations
Optimized the animation loop to update the DOM only when style values change.
Fixed related type issues.


DavidHDev:
LGTM

PR #773, opened 2025-12-08
fix: carousel loop backward navigation animation issue

sukeshhublikar:
Fix bug where swiping backward from cloned first item in loop mode would animate through all items instead of jumping directly to the real last item. Now uses isResetting flag to disable animation and jump directly to the correct position.

Updated all 4 carousel variants:
- content/Components/Carousel/Carousel.jsx
- ts-tailwind/Components/Carousel/Carousel.tsx
- ts-default/Components/Carousel/Carousel.tsx
- tailwind/Components/Carousel/Carousel.jsx


DavidHDev:
I appreciate the effort, but this doesn't seem to have fixed it.

1. It still animates through all the items when you jump from the first item to the last one
2. There are some bigger problems with this component, which causes the loop to completely break:
image

A better refactor is generally needed, I don't think this PR will make a difference. I'll handle it!

PR #776, opened 2025-12-11
Alphabetically sorted component categories for easier navigation

himanshumishra1309:
I was browsing through the React Bits website and noticed that the component list wasn’t in alphabetical order. I found myself taking a little extra time to find what I was looking for, so I thought it might help to sort the categories alphabetically.

It’s not a big change, but from a user’s point of view it does make navigation a bit smoother, especially as the list grows. I’ve gone ahead and opened a PR with the update.

Thanks for taking the time to review it, happy to make any changes if needed!


DavidHDev:
Not requested. Items should not be sorted alphabetically.

DavidHDev:
For easier browsing we have:
- Index Page with filters where items ARE already alphabetically sorted
- Site-wide search functionality if you press "/"

PR #817, opened 2025-12-30
Fix: Resolve React console errors (unique keys and invalid DOM props)

MrQuantum1915:
## Description
This PR resolves two specific React console errors appearing in the browser during development.

### Issues Fixed

1.  **Unique Key Prop Error:**

    * **The Issue:** The `CATEGORIES.map` loop (in `Sidebar.jsx`) was returning a generic Fragment (`<>... </>`), which cannot accept a `key` prop. This caused the "Each child in a list should have a unique 'key' prop" error.
    * **The Fix:** Replaced the shorthand `<>` with react fragment component : `<Fragment key={cat.name}>` to correctly assign the key to the list item wrapper.
    *  [React Fragment](https://react.dev/reference/react/Fragment#:~:text=optional%20key%3A%20Fragments%20declared%20with%20the%20explicit%20%3CFragment%3E%20syntax%20may%20have%20keys.)
  
    * **Screenshot**: 
screenshot_30122025_162845

2.  **Invalid DOM Property Error:**
    * **The Issue:** The `Input` component (in `ComponentList.jsx`) in the search section was receiving both `disabled` and `isDisabled`. The `isDisabled` prop was "leaking" through to the underlying HTML `<input>` element, causing the error: *"React does not recognize the `isDisabled` prop on a DOM element."*
    * **The Fix:** Removed the redundant `isDisabled` prop and kept the standard HTML `disabled` prop. This fixes the error while maintaining the correct disabled behavior and styles.
    * **Screenshot**:
    
screenshot_30122025_173242

## Quality Check
- [x] **Console Check:** Verified that both the "unique key" warning and the "isDisabled" error no longer appear in the browser console.
- [x] **Visual Check:** Confirmed that the Sidebar layout remains unchanged (using `Fragment` preserves the `VStack` spacing).

### DIFF
--- a/src/components/navs/Sidebar.jsx
+++ b/src/components/navs/Sidebar.jsx
@@ -228,8 +228,8 @@ const MainDrawer = ({ isOpen, onClose, categories, location, pendingActivePath, 
                   pendingActivePath={pendingActivePath}
                   handleClick={onClose}
                   handleTransitionNavigation={onNavigation}
-                  onItemMouseEnter={() => { }}
-                  onItemMouseLeave={() => { }}
+                  onItemMouseEnter={() => {}}
+                  onItemMouseLeave={() => {}}
                   itemRefs={{}}
                   isTransitioning={isTransitioning}
                   isFirstCategory={i === 0}


DavidHDev:
Wasn't really bothering me but alright!

Issue #823, opened 2026-01-03
[FEAT]: text-animations/shuffle


Abhxhek:
### Share your suggestion

What i was guessing that the way animation have position left and right, it could have up and down animation too

### Validations

- [ ] I have checked other issues to see if my issue was already discussed or addressed

DavidHDev:
Completed by @Utkarsh-Singhal-26 , closing!

PR #827, opened 2026-01-05
[ FEAT ] : Added Up & Down Support in TextAnimations Shuffle

Utkarsh-Singhal-26:
Added support for the `up` and `down` shuffle direction

Feature as mentioned in issue: https://github.com/DavidHDev/react-bits/issues/823

image


DavidHDev (approved):
I'd say it looks good!

PR #839, opened 2026-01-08
Fix: Render Area Fix for Splash Cursor Animation

amanbobal:
**The splash cursor animation was rendering on the entire screen rather than the canvas area, unlike other animations.**

Tiny Brief of Changes:
- added canvas offset for coords' adjustment
- changed position from 'fixed'->'relative' && overflow to: 'hidden'

---
The Clip below is a Before-After Comparison:

[fix-comparison-vid.webm](https://github.com/user-attachments/assets/7b50a7d9-5d3f-4e54-bb8f-ba3bebe3f2db)

### DIFF
--- a/src/content/Animations/SplashCursor/SplashCursor.jsx
+++ b/src/content/Animations/SplashCursor/SplashCursor.jsx
@@ -962,11 +962,18 @@ function SplashCursor({
       return hash;
     }
 
+    // Get canvas offset for coordinate adjustment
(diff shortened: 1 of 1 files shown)


amanbobal:
PS: You're doing excellent work! Godspeed to you!🫡

DavidHDev:
Hi there, it's like that on purpose, so I don't think this PR is needed. Thank you!

PR #849, opened 2026-01-19
[ FIX ] : Props Table Default Values - Splash Cursor, Gradient Blinds, Waves, Bounce Cards, & Elastic Slider

Utkarsh-Singhal-26:


DavidHDev:
ty!

PR #868, opened 2026-02-02
chore: bump jsrepo

ieedan:
Bumping jsrepo cause we made some changes that slightly change the way dependencies work.

TLDR; we now also resolve dependencies within items aswell. This allows us to handle files being renamed by transforms (more coming soon)

This doesn't actually effect this project though (as far as registry generation) since there aren't any items that actually have _imports_. So just updating in case you have any that do in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "three": "^0.167.1"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


DavidHDev:
Thanks for the bump!

PR #871, opened 2026-02-03
fix: improve type safety in Grainient component

khoinguyenpham04:
Replace 'as any' type assertions with specific types:
- program.uniforms.iTime: { value: number }
- program.uniforms.iResolution: { value: Float32Array }

This improves type safety and prevents potential runtime errors.

Why: I love the new Grainient component and I used it in my codebase, however, after `npm run build` there was a ts error, so I decided to fix it and create this PR. Hope this helps!


Utkarsh-Singhal-26:
Hi @khoinguyenpham04, thanks for the PR! Replacing `as any` with explicit types is a solid improvement 👍

Could you also apply the same change in the ts-default file?
`src/ts-default/Backgrounds/Grainient/Grainient.tsx`

Thanks!

khoinguyenpham04:
Hi @Utkarsh-Singhal-26, thanks for reviewing my PR!

I just applied the same changes to ts-default! Both files now use explicit types instead of as any.

Cheers!

Issue #875, opened 2026-02-05
Check cleanup of WebGL contexts throughout the library


DavidHDev:
It has been reported that some components are not properly cleaned up, leading to high memory usage.

Examples: Hyperspeed, LightPillar

These should be checked and fixed, along with any other components where this applies (if any)

Utkarsh-Singhal-26:
hey @DavidHDev, i am thinking of picking this up, please assign me :relaxed:

PR #882, opened 2026-02-13
Fix Hyperspeed partial options override and improve performance

PedroMarianoAlmeida:
Summary

  - Fix partial effectOptions override — Previously, passing a partial effectOptions prop (e.g. { fov: 120 }) would
  completely discard all other defaults. Defaults are now extracted to a module-level DEFAULT_EFFECT_OPTIONS constant and
  merged at runtime, so users can override individual options without losing the rest.
  - Fix shallow merge on nested colors object — The colors property is now deep-merged separately, so passing { colors: {
  roadColor: 0xff0000 } } only overrides roadColor while preserving all other color defaults.
  - Fix unnecessary re-renders (TS variants) — The useEffect dependency was mergedOptions, a new object created every
  render, causing the effect to re-run constantly. Now depends on the stable effectOptions prop reference.
  - Remove stray console.log — Removed a debug log left inside the render loop (this.options.isHyper).

  Performance note

  Extracting defaults to a module-level constant (DEFAULT_EFFECT_OPTIONS) avoids recreating the object on every render and
   provides a stable reference for useEffect dependencies — but only when no custom prop is passed. When the consumer
  passes an inline effectOptions object, it will still be a new reference on every parent re-render, causing the WebGL
  scene to be destroyed and recreated. A usage tip was added to the docs reminding consumers to memoize their
  effectOptions (via useMemo or a constant) to prevent this.

  Bug fixes
  Bug: Partial options ignored defaults
  Before: <Hyperspeed effectOptions={{ fov: 120 }} /> lost all other defaults
  After: All defaults preserved, only fov overridden
  ────────────────────────────────────────
  Bug: Partial colors replaced entire object
  Before: { colors: { roadColor: 0xff0000 } } wiped other colors
  After: Only roadColor overridden, rest kept
  ────────────────────────────────────────
  Bug: useEffect re-ran every render (TS)
  Before: Dependency on mergedOptions (new object each render)
  After: Dependency on effectOptions (stable ref)
  ────────────────────────────────────────
  Bug: console.log in render loop
  Before: Logged isHyper on every frame
  After: Removed
  Changed files
  
Note: A little related to issue https://github.com/DavidHDev/react-bits/issues/875#issue-3902167451, but does't adress this specific issue (it is an overall performance, but not a cleanup of WebGL contexts

Note 2: If you want, I can record a video running Profiler to see the performance boost

### DIFF
--- a/src/constants/code/Backgrounds/hyperspeedCode.js
+++ b/src/constants/code/Backgrounds/hyperspeedCode.js
@@ -11,6 +11,8 @@ export const hyperspeed = {
 
 // the component will fill the height/width of its parent container, edit the CSS to change this
 // the options below are the default values
+// TIP: if you pass custom effectOptions, memoize the object (e.g. useMemo or a constant)
+// to avoid unnecessary re-renders and WebGL scene recreations
 
 <Hyperspeed
   effectOptions={{


DavidHDev:
Awesome thanks!

PR #884, opened 2026-02-14
Fixes: WebGL Cleanup

Utkarsh-Singhal-26:
Fixes: #875 

Components Affected:
- ASCIIText
- Ballpit
- ColorBends
- FloatingLines
- GhostCursor
- GridDistortion
- GridScan
- Hyperspeed
- LaserFlow
- LightPillar
- LiquidEther
- PixelBlast
- PixelSnow
- PixelTrail
- ShapeBlur


src/ts-tailwind/Animations/ShapeBlur/ShapeBlur.tsx
```diff
@@ -245,12 +244,42 @@ const ShapeBlur: FC<ShapeBlurProps> = ({
...
+    if (materialRef.current) materialRef.current.uniforms.u_circleSize.value = circleSize;
+  }, [circleSize]);
+
+  useEffect(() => {
+    if (materialRef.current) materialRef.current.uniforms.u_circleEdge.value = circleEdge;
+  }, [circleEdge]);
+
+  useEffect(() => {
+    if (materialRef.current) materialRef.current.uniforms.u_pixelRatio.value = pixelRatioProp;
+  }, [pixelRatioProp]);
```
PedroMarianoAlmeida: Why is splitting the changes across several individual useEffects better than before?

Utkarsh-Singhal-26: Splitting the changes across multiple `useEffect` is better because it avoids rebuilding the entire Three.js scene every time a prop changes.

In the original version, any change to `shapeSize`, `roundness`, etc. triggers the main effect, which:
- Disposes the renderer
- Removes the canvas
- Recreates the scene, camera, material, listeners, animation loop

That’s expensive and unnecessary.

DavidHDev:
Hey, I'd like to test this locally before merging but I'm a bit tight on time, so after making these changes, can you confirm:
- Demo sliders in the affected components still work as expected
- Background Studio sliders still work as expected (where applicable)
- No major visual changes in the affected components

Also, it looks like you made no changes to LightPillar, which was one of the components someone actually flagged as problematic, can you verify?

These would be my biggest concerns.

Utkarsh-Singhal-26:
> Hey, I'd like to test this locally before merging but I'm a bit tight on time, so after making these changes, can you confirm:
> 
> * Demo sliders in the affected components still work as expected
> * Background Studio sliders still work as expected (where applicable)
> * No major visual changes in the affected components
> 
> Also, it looks like you made no changes to LightPillar, which was one of the components someone actually flagged as problematic, can you verify?
> 
> These would be my biggest concerns.

Okay @DavidHDev i will check these.

Utkarsh-Singhal-26:
Hi @DavidHDev,

I’ve reviewed the components, and from my side all the necessary cleanup has been completed. This PR looks good to go.

PR #895, opened 2026-02-25
✨ feat(decrypted-text): add click/toggle mode & fix inViewHover re-trigger bug

### LINKED ISSUE #891: [FEAT]: text-animations/decrypted-text: Decrypt text on click
### Share your suggestion

It would be great if you could add an option to the Decrypted Text component to play the animation on click. So the text would be scrambled and once you click on it, it would play the decrypt animation revealing the text and then it would stay decrypted.

### Validations

- [x] I have checked other issues to see if my issue was already discussed or addressed

angelarreolagg:
Fixes #891 
## Summary
This PR adds a `click` mode to the `DecryptedText` component (part of the TextAnimations collection), allowing the animation to trigger when the user clicks on the element. A `clickMode` prop is also introduced, supporting `once` (decrypt on first click, stays decrypted) and `toggle` (alternates between encrypted/decrypted on each click). Since we now have a third animation mode, the `both` option has been renamed to `inViewHover`.

> [!IMPORTANT]
> A pre-existing bug was also fixed in this pass: when using `animateOn="both"`, the hover animation would silently fail after the in-view animation completed. The root cause was both triggers sharing the same `isHovering` boolean — once the view animation set it to `true`, subsequent hover events produced no state change and the effect never re-fired. This was resolved by splitting animation control into dedicated, independent functions (`triggerDecrypt`, `triggerReverse`) backed by more granular state (`isAnimating`, `isDecrypted`, `direction`).

> [!IMPORTANT]
> https://github.com/user-attachments/assets/1e3ca360-7dcc-41db-a753-35f0c6896236

## Changes
- **Click mode added:** `animateOn="click"` starts the text in an encrypted state and decrypts on click.
- **`clickMode` prop added:** `once` (default) decrypts once and stays; `toggle` re-encrypts on the next click.
- **`both` renamed to `inViewHover`:** with a third mode now available, the rename better reflects the actual behavior.
- **Bug fix:** hover animation no longer silently fails after the in-view animation has already run.
- **Code example updated:** `src/constants/code/TextAnimations/decryptedTextCode.js` now includes an example using `animateOn="click"` with `clickMode="toggle"`.

Applied changes in:
- `src/content/TextAnimations/DecryptedText/DecryptedText.jsx`
- `src/tailwind/TextAnimations/DecryptedText/DecryptedText.jsx`
- `src/ts-default/TextAnimations/DecryptedText/DecryptedText.tsx`
- `src/ts-tailwind/TextAnimations/DecryptedText/DecryptedText.tsx`
- `src/constants/code/TextAnimations/decryptedTextCode.js`

## Component Demo — `animateOn` prop updated and `clickMode` control added
- `click` mode added.
- `both` renamed to `View & Hover` with value changed to `inViewHover`.
- new `clickMode` preview control added

Applied changes in:
`src/demo/TextAnimations/DecryptedText.jsx`

## Registry artifacts regenerated
- `public/r/DecryptedText-JS-CSS.json`
- `public/r/DecryptedText-JS-TW.json`
- `public/r/DecryptedText-TS-CSS.json`
- `public/r/DecryptedText-TS-TW.json`

## Demo

https://github.com/user-attachments/assets/2e43bdb4-fe22-4237-849c-0af10c3449a2

## Verification
- `npm run build`: passed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -169,6 +169,7 @@
       "integrity": "sha512-2BCOP7TN8M+gVDj7/ht3hsaO/B/n5oDbiAyyvnRlNOs+u1o+JWNYTQrmpuNp1/Wq2gcFrI01JAW+paEKDMx/CA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.3",
@@ -442,6 +443,7 @@
       "resolved": "https://registry.npmjs.org/@chakra-ui/react/-/react-3.27.0.tgz",
       "integrity": "sha512-M1WTAErI2cYM/PB4h5Kf5CCAg70g3HCzVvTEhcf5ty8QrG6QybPf3RdWSpBlIy7qpjuEnQYpHLxM0jnFLArBgA==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@ark-ui/react": "^5.24.1",
         "@emotion/is-prop-valid": "^1.3.1",
@@ -553,6 +555,7 @@
       "resolved": "https://registry.npmjs.org/@emotion/react/-/react-11.14.0.tgz",
       "integrity": "sha512-O000MLDBDdk/EohJPFUqvnp4qnHeYkVP5B0xEG0D/L7cOKP9kefu2DXn8dj74cQfsEzUqh+sr1RzFqiL1o+PpA==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/runtime": "^7.18.3",
         "@emotion/babel-plugin": "^11.13.5",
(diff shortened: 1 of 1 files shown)


DavidHDev:
That's good, but your code does not do what the issue requested.

The click mode should have the text start out as encrypted, and only decrypted after the first click.

Please read the issue again and update your implementation.

angelarreolagg:
Hey David, thanks for the feedback! English isn't my first language so I just saw "click" and assumed that was the solution — my bad 😅. I re-read the issue and worked on a proper implementation.

While doing so, I noticed that inViewHover (formerly both) had a bug present since the original version, so I fixed that as well — it was actually solved indirectly by the refactor needed to handle the new click mode correctly.

This new version addresses the issue as requested and also ships a clickMode prop with once and toggle support as a small bonus. The PR description has been updated to reflect all the changes. Second time's the charm! 🤞

Issue #222, opened 2025-06-04
[BUG]: text-animations/split-text


skejtfi-studio:
### Describe the issue

SKEJTFI TRAINZ

### Steps to reproduce

SKEJTFI TRAINZ

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

DavidHDev:
Don't open useless issues.

PR #915, opened 2026-03-15
feat: add Hexagons background component

### LINKED ISSUE #854: [FEAT]: Hexagon grid background
### Share your suggestion

Hexagonal grids look **stunning** and I think they would make a great background.
What I have in mind is a similar looking background as [Squares](https://reactbits.dev/backgrounds/squares).

Another option would be to change the Squares background into a Grid background. And add **Shape** customization with Squares or Hexagons (and possibly more) as the variants.

### Validations

- [x] I have checked other issues to see if my issue was already discussed or addressed

aashirpersonal:
## Summary

Extends the existing **Squares** component with a `shape` prop to support multiple grid tile shapes, as discussed in #854. Instead of a separate component, users can now toggle between square and hexagon grids via a single prop.

**Fully backward compatible** — `shape` defaults to `'square'`, so existing usage is unchanged.

## New `shape` Prop

```jsx
<Squares
  shape='hexagon'  // 'square' (default) | 'hexagon'
  speed={0.5}
  squareSize={30}
  direction='diagonal'
  borderColor='#fff'
  hoverFillColor='#222'
/>
```

## What Changed

- **All 4 component variants** (JSX, JSX+Tailwind, TSX, TSX+Tailwind) now accept a `shape` prop
- **Hexagon mode**: flat-top hex grid with proper tiling geometry (`horiz = size × 1.5`, `vert = size × √3`, odd-column offset)
- **Seamless animation** wrapping adapts to hex spacing
- **Hover detection** uses pixel-to-hex coordinate mapping in hexagon mode
- **Demo page** has a new Shape toggle button group (Square / Hexagon)
- **Code example** updated to show the `shape` prop

## Props (updated)

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `direction` | `string` | `'right'` | Animation direction |
| `speed` | `number` | `1` | Animation speed multiplier |
| `borderColor` | `string` | `'#999'` | Border color |
| `squareSize` | `number` | `40` | Size of tiles in pixels |
| `hoverFillColor` | `string` | `'#222'` | Hover fill color |
| `shape` | `string` | `'square'` | **NEW** — `'square'` or `'hexagon'` |

## Files Modified (no new files)

- `src/content/Backgrounds/Squares/Squares.jsx` + CSS variant
- `src/tailwind/Backgrounds/Squares/Squares.jsx`
- `src/ts-default/Backgrounds/Squares/Squares.tsx`
- `src/ts-tailwind/Backgrounds/Squares/Squares.tsx`
- `src/constants/code/Backgrounds/squaresCode.js`
- `src/demo/Backgrounds/SquaresDemo.jsx`
- `src/constants/Information.js` — updated description

Closes #854


DavidHDev:
I was leaning more towards updating/renaming the Squares component to support multiple shapes like the issue states:

```Another option would be to change the Squares background into a ShapeGrid background. And add shape customization with Squares or Hexagons (and possibly more) as the variants.```

aashirpersonal:
That makes a lot more sense! I'll rework this to extend the existing Squares component with a `shape` prop (`'square' | 'hexagon'`) instead of creating a separate component. Will keep backward compatibility — defaulting to `'square'` so nothing breaks. Updating the PR now.

DavidHDev:
cool! I'll merge it to see how it looks :D

PR #928, opened 2026-03-28
Reduce the bundle size of GradualBlur from 60.55 kB gzip to 8.15 kB gzip

Arjunsharmahehe:
The GradualBlur component is a good component, but when I was using it in my project, it exploded the bundle size of my project. I inspected and found that it uses mathjs library.
image

The bundle of GradualBlur in your library is also one of the highest cause of that.
- The default build of the main branch of react-bits
Screenshot 2026-03-28 180558

the 'math' from mathjs is just used in 5 places and to perform only two functions (pow, round) which can be performed by the default "Math" as well. After removing mathjs from being a dependency, and using Math instead of math. The bundle speaks for itsef.
- Build of react-bits after removing mathjs from the component
Screenshot 2026-03-28 180314

A smaller bundle size will only make the sites that use this component feel more snappier and load faster. 

> This PR shows a lot of files changes, that is because I ran `npm run format` and that changed a lot of files T_T. I am sorry for that. I only made the tweaks to all the variants of GradualBlur only, anything else is just a formatting change done by prettier.


DavidHDev:
Please reopen without all the formatting changes. Only GradualBlur should be targeted.

Arjunsharmahehe:
Understandable. Let me get that done

Arjunsharmahehe:
I created a better PR #944  that just changes the 4 essential files. Do check it out.

Issue #937, opened 2026-04-04
[BUG]: text-animations/variable-proximity


tanishworks:
### Describe the issue

The usage uses falloff twice
```  <VariableProximity
    label={'Hover me! And then star React Bits on GitHub, or else...'}
    className={'variable-proximity-demo'}
    fromFontVariationSettings="'wght' 400, 'opsz' 9"
    toFontVariationSettings="'wght' 1000, 'opsz' 40"
    containerRef={containerRef}
    radius={100}
    falloff='linear'
    falloff="linear"  <-- Repeated, remove this
/>
```

### Reproduction Link

https://reactbits.dev/components/variable-proximity

### Steps to reproduce

Go to the site, and look with the eyes :)

### Validations

- [x] I have checked other issues to see if my issue was already reported or addressed

PR #938, opened 2026-04-05
fix(code-example): handle quote variants and dedupe injected props

Utkarsh-Singhal-26:
## Summary
This PR fixes a code-example rendering issue (#937) where the same prop could appear twice in the Usage snippet when quote styles differed.

## Problem
Some usage snippets had an existing prop written with single quotes, while dynamic prop updates are generated with double quotes.  
Because the matcher did not fully account for both styles, it could fail to detect the existing prop and append a new one, causing duplicates.

## Root Cause
The simple prop matcher handled only a subset of value formats and could miss valid single-quoted prop lines.  
When matching failed, fallback logic appended a new prop instead of replacing the existing one.

## Fix
- Expanded prop-line matching to support:
  - Single-quoted strings
  - Double-quoted strings
  - Braced values
  - Bare values
- Added deduplication so if multiple lines for the same prop exist, only one normalized line is kept.

## Why This Works
The injector now recognizes equivalent prop lines regardless of quote style and replaces them in place.  
If duplicates already exist, deduplication removes extras, preventing repeated props in generated Usage output.

## Scope
This change only affects code-example generation.  
There is no runtime behavior change to component logic.

## Validation
- Verified no new errors in the modified file.
- Confirmed duplicate-prop scenarios caused by quote-style mismatch are prevented.

### DIFF
--- a/src/components/code/CodeExample.jsx
+++ b/src/components/code/CodeExample.jsx
@@ -36,13 +36,23 @@ function injectPropsIntoCode(usageCode, props, defaultProps, componentName, demo
     const newPropLine =
       typeof propValue === 'boolean' && propValue === true ? propName : `${propName}=${formattedValue}`;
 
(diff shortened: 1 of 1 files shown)


DavidHDev (commented):
thanks!

PR #944, opened 2026-04-16
fix(GradualBlur): reduce the bundle size from 60.55kB gzip to 8.14kB gzip

Arjunsharmahehe:
Removed the `mathjs` import which was not needed and was just bloating the bundle. It was used just for the `round` and `pow` function which are provided by `Math`. It does not change the behavior of the component and optimizes the bundle a lot. 

Before:
Screenshot 2026-04-16 224541

After:
Screenshot 2026-04-16 224449

### DIFF
--- a/public/r/registry.json
+++ b/public/r/registry.json
@@ -528,9 +528,7 @@
 			"title": "GradualBlur",
 			"description": "Progressively un-blurs content based on scroll or trigger creating a cinematic reveal.",
 			"type": "registry:component",
-			"dependencies": [
-				"mathjs@^14.6.0"
-			],
+			"dependencies": [],
 			"registryDependencies": [],
 			"files": [
 				{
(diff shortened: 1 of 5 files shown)


Utkarsh-Singhal-26:
Also remove the unnecessary `mathjs` dependency from `public/r/registry.json` for the `GradualBlur` component.
Updated:

```json
"dependencies": []
```

(previously included `"mathjs@^14.6.0"`)

Arjunsharmahehe:
I removed the dependency from the `public/r/registry.json`, there appear to be some other changes in the diff view but that is just a rearrangement in the order of listing of the element. of the Hyperspeed component.  It appears to have happened on build. It should not cause any issue so that is fine I suppose.

DavidHDev (approved):

PR #963, opened 2026-05-14
fix: resync canvas size on browser zoom change in ElectricBorder

### LINKED ISSUE #962: [BUG]: Electric Border breaks on browser zoom
### Describe the issue

Hi! I found an issue with the Electric Border animation component.

When changing the browser zoom level (zoom in/out),
the electric border effect becomes misaligned and extends outside the rectangular border.

Steps to reproduce:
1. Render the Electric Border component
2. Change the browser zoom level (e.g. 80%, 125%)
3. Observe the border effect positioning

Expected behavior:
The electric border should stay aligned with the rectangle border regardless of browser zoom level.

Actual behavior:
The electric effect shifts outside the border area after zooming.

Environment:
- Browser: Chrome
- React: 18.x
- OS: Windows 11

### Reproduction Link

_No response_

### Steps to reproduce

Steps to reproduce:
1. Open a page using the Electric Border component
2. Change the browser zoom level (e.g. 125% or 80%)
3. Observe the electric border effect
4. The animation becomes misaligned and extends outside the rectangle

### Validations

KhaledSaeed18:
## Fix: Electric Border misalignment on browser zoom

Fixes #962

## Problem

When changing browser zoom level (e.g. 80%, 90%), `window.devicePixelRatio` changes but the `ResizeObserver` on the container does not fire. CSS layout dimensions are unchanged. This left the canvas buffer sized for the old DPR while the draw loop applied `ctx.scale` with the new one, causing the electric border stroke to shift outside the rectangle.

Affected variants: Button, Card, and Circle (all four JS/CSS, JS/TW, TS/CSS, TS/TW).

## Solution

Track `lastDpr` as a closure variable inside the `useEffect`. At the top of every `drawElectricBorder` frame, read the current DPR and compare. If it differs, call `updateSize()` immediately before drawing to resize the canvas buffer and sync `width`/`height`. The check is a single comparison per frame no overhead on normal renders.

## Changes

- `src/content/Animations/ElectricBorder/ElectricBorder.jsx`
- `src/tailwind/Animations/ElectricBorder/ElectricBorder.jsx`
- `src/ts-default/Animations/ElectricBorder/ElectricBorder.tsx`
- `src/ts-tailwind/Animations/ElectricBorder/ElectricBorder.tsx`
- `public/r/ElectricBorder-*.json` (auto-generated registry mirrors)

## Before / After

| | Before | After |
|---|---|---|
| **75% zoom** | eb-75-b | eb-75-a |
| **80% zoom** | eb-80-b | eb-80-a |

## Testing

- [x] Rendered all ElectricBorder variants (Button, Card, Circle)
- [x] Tested browser zoom at 75%, 80%, 90%, 100%, 110%, 120% ...
- [x] Border stays aligned with the rectangle at all zoom levels
- [x] No visual regression at default zoom (100%)
- [x] No performance or side-effect issues observed during animation


DavidHDev:
Thanks for the fix!

PR #971, opened 2026-05-27
fix(OrbitImages): prevent responsive scale flash on mount and refresh

### LINKED ISSUE #970: [BUG]: OrbitImages flashes oversized on mount/refresh when responsive is enabled
### Describe the issue

When `responsive={true}`, OrbitImages briefly renders at full design size (`scale=1`) before shrinking to fit the container. This happens on initial page load and when remounting the component (e.g. clicking Refresh on the demo page).
The visual effect looks like a quick zoom-in then zoom-out.

### Expected behavior

The component should render at the correct scale immediately, with no visible size flash.

### Root cause

`scale` is initialized to `1` and measured in `useEffect` (after paint). Before measurement completes, the 1400px design canvas is shown at `scale(1)`.

### Reproduction Link

_No response_

### Steps to reproduce

1. Open `/animations/orbit-images` (demo uses `responsive={true}`)
2. Click the “Refresh” button, reload the page (press F5), or adjust the parameters.
3. Observe the orbit briefly appearing too large, then snapping to the correct size

### Validations

achelie:
## Summary

Fix OrbitImages briefly rendering at the wrong size when `responsive={true}`.
On mount and remount (e.g. demo Refresh), the component initially used `scale=1` before `ResizeObserver` updated it to the correct value, causing a visible zoom-in-then-shrink flash.

## Solution:
- Initializes `scale` as `null` instead of `1`
- Measures container scale in `useLayoutEffect` (before paint)
- Hides the scaling layer with `visibility: hidden` until scale is ready
- Applies transform only when `scale !== null`
- Uses `pathWidth / (scale ?? 1)` for safe stroke width in non-responsive mode
- Updates all 4 variants (JS-CSS, JS-TW, TS-CSS, TS-TW) and registry JSON files

Fixes #970

**Before (bug):**

[**After (fix):**](https://github.com/user-attachments/assets/5c5a6385-7533-4621-b764-f60b009a8f20)

**After (fix):**

[> Tip: Drag your `.mp4` files into this comment box and replace the placeholders above with the generated GitHub links.](https://github.com/user-attachments/assets/8415e1b0-3de4-4468-bf7c-ff604f4e62c0)

## Test plan

- [x] Open `/animations/orbit-images` locally
- [x] Click **Refresh** multiple times — no zoom-in/zoom-out flash
- [x] Reload the page (F5) — same result
- [x] Verified on desktop; no console errors
- [x] All 4 component variants updated
- [x] Registry outputs for OrbitImages regenerated


DavidHDev:
Thanks for the fix!

PR #1001, opened 2026-07-14
darkveil fixes

urliveyn:


urliveyn:
I added a fix to prevent the effect from shrinking on mobile devices; I can't remove the spaces because of IntelliJ IDEA.

urliveyn:
@DavidHDev, you asked for corrections, and I've done everything!

PR #1006, opened 2026-07-21
fix: guard render-time window/matchMedia access for SSR (DecayCard, PixelCard, Masonry)

zbysir:
## Problem

`DecayCard`, `PixelCard` and `Masonry` access browser globals **during render** (in `useRef` / `useState` initializers), so they throw under SSR (e.g. Next.js, or any `renderToString` environment) before the component ever mounts:

```
ReferenceError: window is not defined
```

Same class of bug fixed for `TargetCursor` in #911; I've reused the same `typeof window` guard style.

## Changes

Guarded only the **render-time** accesses — effect/handler code is untouched.

- **DecayCard** — the `cursor` and `winsize` `useRef` initializers read `window.innerWidth/innerHeight`. Guarded with `typeof window !== 'undefined' ? … : 0`. The existing resize/mousemove effect overwrites these on mount, so client behavior is unchanged.
- **PixelCard** — `reducedMotion = useRef(window.matchMedia(...).matches)` runs at render. Guarded with `typeof window !== 'undefined' && …` (→ `false` on the server, i.e. motion not reduced).
- **Masonry** — `useMedia`'s `get()` is passed to `useState(get)`, so it runs once during render and calls `matchMedia`. Added an early `if (typeof window === 'undefined') return defaultValue;`.

All three were updated across **all 4 variants** (JS-CSS, JS-TW, TS-CSS, TS-TW) per the contributing guide.

## Notes / testing

- No API or client-behavior changes; server-render safety only.
- Verified the changed files parse and are Prettier-clean (`prettier --check`).
- Scope kept tight: only components with **unguarded render-time** access. Others referencing `window` do so inside effects/handlers or already guard with `typeof window`, so they're SSR-safe and left as-is.

### DIFF
--- a/src/content/Components/DecayCard/DecayCard.jsx
+++ b/src/content/Components/DecayCard/DecayCard.jsx
@@ -16,9 +16,15 @@ const DecayCard = ({
 }) => {
   const svgRef = useRef(null);
   const displacementMapRef = useRef(null);
-  const cursor = useRef({ x: window.innerWidth / 2, y: window.innerHeight / 2 });
+  const cursor = useRef({
+    x: typeof window !== 'undefined' ? window.innerWidth / 2 : 0,
+    y: typeof window !== 'undefined' ? window.innerHeight / 2 : 0
+  });
   const cachedCursor = useRef({ ...cursor.current });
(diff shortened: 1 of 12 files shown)


DavidHDev:
nice, thanks!

PR #1015, opened 2026-07-29
fix: restart animation loop when stale requestAnimationFrame exists

### LINKED ISSUE #1014: [BUG] Animation loop can become permanently stuck because startLoop() never reschedules when rafRef.current is stale
https://github.com/DavidHDev/react-bits/blob/61909958fab8f90141f61b449bcb5696ba2c78b3/src/content/Components/LineSidebar/LineSidebar.jsx#L87
https://github.com/DavidHDev/react-bits/blob/61909958fab8f90141f61b449bcb5696ba2c78b3/src/content/Components/OptionWheel/OptionWheel.jsx#L131

### Component(s)

- OptionWheel
- LineSidebar

Possibly other components using the same animation loop pattern.

### Description

Both components can stop animating because `startLoop()` exits early when `rafRef.current` is already set.

Current implementation:

```ts
const startLoop = useCallback(() => {
  if (rafRef.current != null) return;

  lastRef.current = performance.now();
  rafRef.current = requestAnimationFrame(runFrame);
}, [runFrame]);
```

If `rafRef.current` remains non-null while no animation callback is running, all future calls to `startLoop()` return immediately.

As a result:

- OptionWheel never applies transforms
- LineSidebar never updates `--effect`
- Hover animations don't work
- Selection animations don't start

### Expected

Calling `startLoop()` should always guarantee an animation loop is running.

### Actual

The component can enter a permanently idle state where `rafRef.current` contains an old requestAnimationFrame id but no callback executes.

### Environment

- React 18.2
- Vite 5
- Installed using the official shadcn CLI

### Workaround

Replacing the guard with a restart fixes the issue:

```ts
if (rafRef.current != null) {
    cancelAnimationFrame(rafRef.current);
}

lastRef.current = performance.now();
rafRef.current = requestAnimationFrame(runFrame);
```

I verified this fixes both OptionWheel and LineSidebar.

ishubhamrathi:
## Summary

This PR fixes an animation loop issue affecting multiple components that use the same `requestAnimationFrame` scheduling pattern.

## Root Cause

The animation loop was guarded with:

```ts
if (rafRef.current != null) return;
```

If `rafRef.current` contained a previously scheduled animation frame ID, `startLoop()` exited early without scheduling a new frame. This could prevent the animation loop from restarting, causing interactive animations to stop responding.

## Fix

Restart the animation loop by cancelling any previously scheduled frame before requesting a new one.

```ts
const startLoop = useCallback(() => {
  if (rafRef.current != null) {
    cancelAnimationFrame(rafRef.current);
  }

  lastRef.current = performance.now();
  rafRef.current = requestAnimationFrame(runFrame);
}, [runFrame]);
```

## Components Updated

- OptionWheel
- LineSidebar

## Testing

Verified that the affected components correctly restart their animation loop and interactive animations (hover, selection, etc.) now work as expected.

Fixes #1014

### DIFF
--- a/src/content/Components/LineSidebar/LineSidebar.jsx
+++ b/src/content/Components/LineSidebar/LineSidebar.jsx
@@ -84,7 +84,10 @@ const LineSidebar = ({
   }, []);
 
   const startLoop = useCallback(() => {
-    if (rafRef.current != null) return;
+    if (rafRef.current != null) {
+      cancelAnimationFrame(rafRef.current);
+    }
+
     lastRef.current = performance.now();
     rafRef.current = requestAnimationFrame(runFrame);
   }, [runFrame]);
--- a/src/content/Components/OptionWheel/OptionWheel.jsx
+++ b/src/content/Components/OptionWheel/OptionWheel.jsx
@@ -128,7 +128,9 @@ const OptionWheel = ({
   }, []);
 
   const startLoop = useCallback(() => {
-    if (rafRef.current != null) return;
+    if (rafRef.current != null) {
+      cancelAnimationFrame(rafRef.current);
+    }
     lastRef.current = performance.now();
     rafRef.current = requestAnimationFrame(runFrame);
   }, [runFrame]);
--- a/src/tailwind/Components/LineSidebar/LineSidebar.jsx
+++ b/src/tailwind/Components/LineSidebar/LineSidebar.jsx
@@ -83,7 +83,9 @@ const LineSidebar = ({
   }, []);
 
   const startLoop = useCallback(() => {
-    if (rafRef.current != null) return;
+    if (rafRef.current != null) {
+      cancelAnimationFrame(rafRef.current);
+    }
     lastRef.current = performance.now();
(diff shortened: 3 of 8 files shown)


DavidHDev:
Have you checked the other 3 variants of each of these affected components? The change should be made everywhere.

ishubhamrathi:
> Have you checked the other 3 variants of each of these affected components? The change should be made everywhere.

I verified the other three variants while reproducing the issue. They use different implementations and don't exhibit this behavior, so this fix is only needed for OptionWheel and LineSidebar.

DavidHDev:
@ishubhamrathi what I mean is, OptionWheel and LineSidebar both have 3 other files where the code is esentially the same. The library has 4 variants to each component (JS+CSS, JS+TAILWIND, TS+CSS, TS+TAILWIND). You only modified the JS+CSS versions in this PR, when all 3 others must be kept in sync.

ishubhamrathi:
@DavidHDev Thanks for the review. I've updated the remaining component variants as well.

DavidHDev (approved):

PR #1018, opened 2026-07-31
fix(ts): use type-only imports in TS variants (fixes TS1484 in stock Vite react-ts apps)

noron12234:
## The problem

React Bits components are meant to be copied into a consumer's project. A stock
`npm create vite@latest -- --template react-ts` project ships
`verbatimModuleSyntax: true`, and under that flag every TS variant that imports
a type through a value import fails to compile.

Reproduction, against `main` as it is today:

```bash
npm create vite@latest repro -- --template react-ts
cd repro && npm i
curl -o src/Magnet.tsx https://raw.githubusercontent.com/DavidHDev/react-bits/main/src/ts-default/Animations/Magnet/Magnet.tsx
npx tsc --noEmit
```

```
src/Magnet.tsx(1,46): error TS1484: 'ReactNode' is a type and must be imported
  using a type-only import when 'verbatimModuleSyntax' is enabled.
src/Magnet.tsx(1,57): error TS1484: 'HTMLAttributes' is a type and must be
  imported using a type-only import when 'verbatimModuleSyntax' is enabled.
```

With this branch, the same check exits 0.

## Why CI is green on `main` even though this is broken

This repo's own `tsconfig.json` sets `isolatedModules` but **not**
`verbatimModuleSyntax`, so `tsc` here never raises TS1484. The breakage only
appears downstream, in the consumer project — which is the only place these
files are ever compiled. That is why there is no failing test to point at.

If you would rather catch this in CI going forward, adding
`"verbatimModuleSyntax": true` to `tsconfig.json` would do it. I left that out
of this PR deliberately since it changes your build config.

## What this changes

83 `.tsx` files, mechanically: type-only specifiers (`type Foo`) for imports
that are types. `+126 / -94`. No runtime change — `verbatimModuleSyntax`
governs emit of import statements only.

Generated `public/r/**` artifacts are **not** included; `npm run build` runs
`registry:build` and regenerates them.

Precedent: magicui merged the identical fix for the same root cause in
magicuidesign/magicui#993.

## Reviewing this

The diff is wide but shallow — one line per file in most cases. The mechanical
check is that every added `type` keyword sits in front of an identifier that is
only ever used in a type position.

### DIFF
--- a/src/ts-default/Animations/Crosshair/Crosshair.tsx
+++ b/src/ts-default/Animations/Crosshair/Crosshair.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, RefObject } from 'react';
+import React, { useEffect, useRef, type RefObject } from 'react';
 import { gsap } from 'gsap';
 
 const lerp = (a: number, b: number, n: number): number => (1 - n) * a + n * b;
--- a/src/ts-default/Animations/ElectricBorder/ElectricBorder.tsx
+++ b/src/ts-default/Animations/ElectricBorder/ElectricBorder.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useCallback, CSSProperties, ReactNode } from 'react';
+import React, { useEffect, useRef, useCallback, type CSSProperties, type ReactNode } from 'react';
 import './ElectricBorder.css';
 
 interface ElectricBorderProps {
--- a/src/ts-default/Animations/GradualBlur/GradualBlur.tsx
+++ b/src/ts-default/Animations/GradualBlur/GradualBlur.tsx
@@ -1,4 +1,4 @@
-import React, { CSSProperties, useEffect, useRef, useState, useMemo, PropsWithChildren } from 'react';
+import React, { type CSSProperties, useEffect, useRef, useState, useMemo, type PropsWithChildren } from 'react';
 
 import './GradualBlur.css';
 
--- a/src/ts-default/Animations/ImageTrail/ImageTrail.tsx
+++ b/src/ts-default/Animations/ImageTrail/ImageTrail.tsx
@@ -1,5 +1,5 @@
 import { gsap } from 'gsap';
-import { JSX, useEffect, useRef } from 'react';
+import { type JSX, useEffect, useRef } from 'react';
 import './ImageTrail.css';
 
 function lerp(a: number, b: number, n: number): number {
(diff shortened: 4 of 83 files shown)


noron12234:
Follow-up: I understated the impact in the description. For at least one component this is not a `tsc` complaint at all — it breaks the bundler and the dev server, with TypeScript entirely out of the picture.

`Ballpit` imports `WebGLRendererParameters` from `three`, which is a type. `three` does not export it at runtime, and unlike the other cases the name survives into the emitted module.

**Production build**, stock Vite React-TS app, `tsc` skipped so only the bundler runs:

```console
$ npx vite build
error during build:
Build failed with 1 error:
[MISSING_EXPORT] "WebGLRendererParameters" is not exported by "node_modules/three/build/three.module.js".
 4 │ import { ACESFilmicToneMapping, ..., WebGLRenderer, WebGLRendererParameters } from "three";
   │                                                     ╰───────────── Missing export
```

**Dev server**, same app, measured with Playwright:

| | page error | component renders |
|---|---|---|
| current `main` | `The requested module '/node_modules/.vite/deps/three.js' does not provide an export named 'WebGLRendererParameters'` | **no — blank page** |
| this PR | none | yes |

So `Ballpit`'s two TS variants cannot be built or run at all in a stock Vite app today, regardless of anyone's `tsconfig`.

Why this one and not `Silk`: in `Silk`, `IUniform` is its own single-specifier `import { IUniform } from 'three'`, and an import statement whose only specifier is unused after type erasure is dropped before it reaches the bundler. `Ballpit`'s sits inside a 21-name import whose other names are genuine values, so the statement survives — and carries the type name with it. That is why the failure looks inconsistent across components rather than universal.

Nothing changes in the diff; the `type` modifier this PR adds at `Ballpit.tsx:25` is what fixes it. Just wanted the severity on the record, since "compile error" undersells a component that currently fails `vite build` outright.

noron12234:
Pushed `6ad9ba0`: the registry artifacts were missing from this PR.

`public/r/*.json` embeds the component source verbatim and is what the jsrepo CLI writes into a user's project, so fixing only `src/` left the shipped payload with the value imports intact. That includes `Ballpit`'s `WebGLRendererParameters`, which is the one that makes rolldown fail the build outright rather than just upsetting `tsc`.

Regenerated with `npm run registry:build`; 83 entries change, all TS variants.

noron12234:
Slimmed this down and rewrote the description with a reproduction.

Two things changed since you last would have seen it:

- The generated `public/r/**` artifacts are gone — the diff is source only now. That was 83 of the ~166 files and it was pure noise, `npm run build` regenerates them.
- The description now has a copy-pasteable repro showing `tsc` failing on `main` in a stock Vite `react-ts` project, and passing on this branch.

Worth calling out explicitly: this repo's own `tsconfig.json` doesn't set `verbatimModuleSyntax`, so your CI genuinely cannot catch this — the break only shows up in the consumer's project after they copy a component. That's why there's no failing test here to point at.

Happy to split it per-category if a single 83-file diff is awkward to review.

DavidHDev (approved):

PR #1022, opened 2026-07-31
fix(Lightning, RippleGrid): cancel the render loop on cleanup

noron12234:
## The problem

`Lightning` and `RippleGrid` each run a self-scheduling `requestAnimationFrame` loop and never cancel it:

```jsx
const render = () => {
  ...
  gl.drawArrays(gl.TRIANGLES, 0, 6);
  requestAnimationFrame(render);      // schedules itself forever
};
requestAnimationFrame(render);

return () => {
  window.removeEventListener('resize', resizeCanvas);   // the loop is not touched
};
```

So the loop outlives the component. `RippleGrid` makes it starker — its cleanup drops the WebGL context and removes the canvas, and the loop then keeps calling `renderer.render(...)` on a context that has been deliberately lost.

`Lightning` has a second problem on top. Its effect depends on `[hue, xOffset, speed, intensity, size]`, so **every prop change re-runs the effect and starts another loop** without stopping the previous one.

## Measurement

`Lightning` in a stock Vite React app, with `requestAnimationFrame` wrapped to count callbacks per second:

| | current `main` | this PR |
|---|---|---|
| 1 mount, no prop changes | ~60/s | ~60/s |
| after changing `hue` 4 times | **~300/s** (5 loops) | ~60/s |
| **after unmounting** | **~305/s** (all 5 still running) | **~0/s** |

The middle row is the prop-change pile-up; the last row is the leak. On `main` the component is gone from the tree and five render loops are still driving WebGL every frame.

`RippleGrid` has the identical loop shape and cleanup omission — I verified that by reading all four variants, but I could not get it to initialise inside my minimal probe harness, so I am **not** claiming a measured number for it.

## The change

Capture the frame id and cancel it in the cleanup:

```jsx
let animationFrameId;
const render = () => {
  ...
  animationFrameId = requestAnimationFrame(render);
};
animationFrameId = requestAnimationFrame(render);

return () => {
  cancelAnimationFrame(animationFrameId);
  window.removeEventListener('resize', resizeCanvas);
};
```

Nothing else changes. On `Lightning` this also means a prop change now cancels the old loop before the effect re-runs, which is what the existing cleanup was always supposed to do.

## Scope

Both components, all four variants — 8 files, +32/-16.

## Verification

- The measurement above, before and after.
- `npx tsc --noEmit`: no errors mentioning either component.
- `npx prettier --check` clean on all eight files.
- `npx vite build` passes.

Found by scanning effects that create a resource without releasing it in the same effect. Three more components (`SplashCursor`, `FallingText`, `Crosshair`) show the same shape; I left them out of this PR so the diff stays to the two I could reason about fully, and can follow up if you want them.

### DIFF
--- a/src/content/Backgrounds/Lightning/Lightning.jsx
+++ b/src/content/Backgrounds/Lightning/Lightning.jsx
@@ -154,6 +154,7 @@ const Lightning = ({ hue = 230, xOffset = 0, speed = 1, intensity = 1, size = 1 
     const uIntensityLocation = gl.getUniformLocation(program, 'uIntensity');
     const uSizeLocation = gl.getUniformLocation(program, 'uSize');
 
+    let animationFrameId;
     const startTime = performance.now();
     const render = () => {
       resizeCanvas();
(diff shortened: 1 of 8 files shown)


noron12234:
Pushed the rebuilt registry artifacts. `public/r/*.json` embeds the component source verbatim and is what the jsrepo CLI writes into a user's project, so the source-only fix still shipped the broken version. Regenerated with `npm run registry:build`.

DavidHDev (approved):

PR #1046, opened 2026-08-15
fix(ts): resolve the 14 strict typecheck errors in the TS variants

### LINKED ISSUE #1041: [BUG]:  TypeScript variants fail strict typecheck with 14 errors
### Describe the issue

After a clean installation from the current `main` branch, running TypeScript in strict mode reports 14 errors across six TS-CSS and TS-TW variant files.

  Affected components:

  - AcidSquares (`ts-default` and `ts-tailwind`)
  - SpecularButton (`ts-default` and `ts-tailwind`)
  - SplitText (`ts-default` and `ts-tailwind`)

  The repository does not currently provide a `typecheck` script, so I ran the installed TypeScript compiler directly using the repository's existing
  `tsconfig.json`.

  ### Expected behavior

  All TypeScript component variants should compile successfully with the repository's strict TypeScript configuration.

  ### Actual behavior

  TypeScript exits with 14 errors, including:

  - `RenderTarget | null` being passed where `RenderTarget | undefined` is expected
  - A possibly-null `target` being accessed
  - The polymorphic SplitText element resolving its props and children to `never`

  ### Environment

  - Commit: `c7109dccb42e06592d1d9bc50bc87204697240e2`
  - Branch: `main`
  - Node.js: `v24.19.0`
  - npm: `11.17.0`
  - OS: Windows 11 (`10.0.26100`)

### Reproduction Link

_No response_

### Steps to reproduce

 1. Clone the repository and check out the affected commit:

     ```bash
     git clone https://github.com/DavidHDev/react-bits.git
     cd react-bits
     git checkout c7109dccb42e06592d1d9bc50bc87204697240e2

  2. Install dependencies from the lockfile:

     npm ci

  3. Run TypeScript using the repository configuration:

     npm exec tsc -- --noEmit

  4. TypeScript reports 14 errors:

     src/ts-default/Backgrounds/AcidSquares/AcidSquares.tsx(298,40): error TS2322: Type 'RenderTarget | null' is not assignable to type 'RenderTarget |
     undefined'.
     src/ts-default/Backgrounds/AcidSquares/AcidSquares.tsx(304,44): error TS2322: Type 'RenderTarget | null' is not assignable to type 'RenderTarget |
     undefined'.

     src/ts-default/Components/SpecularButton/SpecularButton.tsx(236,22): error TS18047: 'target' is possibly 'null'.

     src/ts-default/TextAnimations/SplitText/SplitText.tsx(171,8): error TS2745: This JSX tag's 'children' prop expects type 'never' which requires multiple
     children, but only a single child was provided.
     src/ts-default/TextAnimations/SplitText/SplitText.tsx(171,12): error TS2322: Type 'RefObject<HTMLParagraphElement | null>' is not assignable to type
     'never'.
     src/ts-default/TextAnimations/SplitText/SplitText.tsx(171,22): error TS2322: Type 'CSSProperties' is not assignable to type 'never'.
     src/ts-default/TextAnimations/SplitText/SplitText.tsx(171,36): error TS2322: Type 'string' is not assignable to type 'never'.

     src/ts-tailwind/Backgrounds/AcidSquares/AcidSquares.tsx(297,40): error TS2322: Type 'RenderTarget | null' is not assignable to type 'RenderTarget |
     undefined'.
     src/ts-tailwind/Backgrounds/AcidSquares/AcidSquares.tsx(303,44): error TS2322: Type 'RenderTarget | null' is not assignable to type 'RenderTarget |
     undefined'.

     src/ts-tailwind/Components/SpecularButton/SpecularButton.tsx(241,22): error TS18047: 'target' is possibly 'null'.

     src/ts-tailwind/TextAnimations/SplitText/SplitText.tsx(168,8): error TS2745: This JSX tag's 'children' prop expects type 'never' which requires
     multiple children, but only a single child was provided.
     src/ts-tailwind/TextAnimations/SplitText/SplitText.tsx(168,12): error TS2322: Type 'RefObject<HTMLParagraphElement | null>' is not assignable to type
     'never'.
     src/ts-tailwind/TextAnimations/SplitText/SplitText.tsx(168,22): error TS2322: Type 'CSSProperties' is not assignable to type 'never'.
     src/ts-tailwind/TextAnimations/SplitText/SplitText.tsx(168,36): error TS2322: Type 'string' is not assignable to type 'never'.

  Expected result: the command exits successfully without TypeScript errors.

### Validations

aron-intframe:
Closes #1041.

`npx tsc --noEmit` gives 14 errors on main, this takes it to 0. Three unrelated causes:

**AcidSquares** (4 errors) — `rtA`/`rtB` are `RenderTarget | null` and `ensureTargets()` fills them in, but tsc can't see through the call. The two lines right below each `render()` already use `rtA!`/`rtB!`, so the render calls were the odd ones out. Made them match.

**SpecularButton** (2) — the null check goes through an intermediate `steer` boolean, and narrowing doesn't survive that. Inlined the condition into the ternary. Same operands in the same order, so nothing changes at runtime.

**SplitText** (8) — this one's the cast itself, which is why I called it out. `tag` is already typed `'h1' | ... | 'span'`, and `as React.ElementType` widens that to every intrinsic element. JSX intersects props across a union, so with `input`, `svg` and friends in there, `children`/`ref`/`style`/`className` all collapse to `never`. That's exactly the 4 errors per file. Dropping the cast fixes it, and the type ends up more precise than the cast was.

Ran `npx vite build` after, still builds (13.3s).

Two things I deliberately left alone:

- The JS variants don't need matching changes. All 14 are type-only — `ensureTargets()` does populate the targets, `steer` does include the null check, and the 8 tags do accept those props. Mentioning it because of the four-variants rule in CONTRIBUTING.
- `prettier --check` already fails on both `SpecularButton.tsx` files on main. Running `--write` pulls a bunch of unrelated reformatting into the diff, so I left it.

AI was used for assistance.


DavidHDev:
Landed in #1053 with contributor authorship preserved. Strict TypeScript now passes; AcidSquares had already been fixed on current main. Thank you!

PR #1047, opened 2026-08-15
fix: resolve no-empty lint errors in WarpText

houssamb4:
This PR fixes ESLint no-empty errors in the WarpText component by replacing empty catch blocks with explicit no-op statements


DavidHDev:
Thanks for the cleanup. This is not mergeable yet because it only changes the JS-CSS WarpText variant: the same empty catches remain in JS-TW, TS-CSS, and TS-TW, and ESLint still fails on JS-TW. Please mirror the fix across all four variants and run the targeted lint check.

PR #1048, opened 2026-08-16
Fix Scroll Stack layout on small screens

EnderRomantice:
I've noticed that this component has some display issues on smaller screens. I'm using the latest version of Chrome on a 13-inch MacBook Air M5.
截屏2026-08-16 10 37 58

I fix that

https://github.com/user-attachments/assets/7e586e16-5f63-447f-90ab-735f96f9ab43


DavidHDev:
Landed in #1053 with contributor authorship preserved. I also verified the responsive row/column behavior at multiple viewport widths with no horizontal overflow. Thank you!

PR #1050, opened 2026-08-22
fix(registry): ship stylesheets as registry:file so `shadcn add` works

### LINKED ISSUE #1049: [BUG]: Error: "Unexpected token (1:0)" when trying to install @react-bits/Lanyard-JS-CSS via shadcn CLI
### Describe the issue

I'm encountering a persistent error when attempting to add the @react-bits/Lanyard-JS-CSS component using the shadcn CLI. The installation fails with an "Unexpected token (1:0)" error across multiple versions of the CLI.

### Reproduction Link

_No response_

### Steps to reproduce

Run the following command in a Next.js/React project:

bash
npx shadcn@4.18.0 add @react-bits/Lanyard-JS-CSS
Observe the error:

text
✔ Checking registry.
⠼ Updating files.

Something went wrong. Please check the error below for more details.
If the problem persists, please open an issue on GitHub.

Unexpected token (1:0)
Tried downgrading to previous versions as suggested:

bash
npx shadcn@4.17.0 add @react-bits/Lanyard-JS-CSS
npx shadcn@4.16.0 add @react-bits/Lanyard-JS-CSS
All attempts resulted in the same error.

Expected Behavior
The component should install successfully via the shadcn CLI.

Actual Behavior
Installation fails with "Unexpected token (1:0)" error. The process gets stuck at "Updating files" before failing.

Environment
OS: Windows (PowerShell)

Node Version: [Add your Node version]

Package Manager: npm

Project: Next.js/React portfolio project

shadcn CLI Versions Attempted: 4.18.0, 4.17.0, 4.16.0

Additional Context
Running in PowerShell (PS C:\Users\eyob\portfolio>)

The CLI successfully checks the registry but fails during file updates

The error message suggests it might be a parsing issue with the component files

Possible Cause
The "Unexpected token (1:0)" error typically indicates a syntax error at the first character of a file. This could be due to:

Invalid JSON/JSX in the component source

Malformed registry entry

Encoding issues with the component files

Troubleshooting Already Tried
✅ Multiple CLI versions (4.18.0, 4.17.0, 4.16.0)

✅ Fresh installation attempts

✅ npx shadcn@latest with and without version pinning

Question
Is there a specific compatibility requirement or known issue with the @react-bits/Lanyard-JS-CSS component? The component name suggests it might be a CSS/JS hybrid, which could be causing parsing issues.

### Validations

Agnik47:
Closes #1049

## Problem

`npx shadcn add @react-bits/<Component>-JS-CSS` fails with `Unexpected token (1:0)` right after "Checking registry", at the "Updating files" step.

It isn't specific to Lanyard — **every one of the 292 `-CSS` registry entries** is affected. `-TW` variants ship no stylesheet and were always fine.

The shadcn CLI runs every file it installs through its JavaScript/TypeScript transformer unless the file is typed `registry:file`. `jsrepo.config.ts` shipped each component as a whole folder, and a folder can't declare a type per file, so stylesheets went out as `registry:component` and got parsed as JavaScript. `Unexpected token (1:0)` is the parser choking on the first character of the CSS.

## Fix

Stylesheets are now declared as `registry:file`, which shadcn copies verbatim. `registry:file` requires an explicit target, so it is set to `@components/<Name>.css` — shadcn resolves the `@components` alias to whatever the user configured in their `components.json`, which is where the component itself lands, so the component's relative `./<Name>.css` import keeps resolving.

Because of the per-file type, variants that actually ship a stylesheet are now listed file by file instead of as a folder. Everything else still ships as a folder, so all `-TW` entries stay byte-identical.

`public/r/` is regenerated (`npm run registry:build`). That rebuild also picks up ~57 entries whose committed content had drifted from `src/` before this PR — those are content-only refreshes, no structural change.

## Testing

- Reproduced first against the live registry with shadcn `4.16.0`, `4.17.0` and `4.18.0` — all three fail on `Lanyard-JS-CSS`, `Aurora-JS-CSS` and `AccordionGallery-JS-CSS`, while `Aurora-JS-TW` installs fine.
- Re-ran `npx shadcn add` against the rebuilt registry on all three CLI versions, in both a JS project (`"tsx": false`) and a TS project (`"tsx": true`): `Lanyard`, `Aurora`, `Hyperspeed`, `AccordionGallery`, `Magnet`, `BlurText`, `ASCIIText` (`-JS-CSS` / `-TS-CSS`) plus `Aurora-JS-TW` and `SplitText-TS-TW` — all install cleanly. Stylesheets land next to their component and every relative import resolves; `Hyperspeed` still gets `HyperSpeedPresets` alongside it.
- Asserted across all 664 generated items: 292 CSS entries, each `registry:file` with a matching `@components/<Name>.css` target; every other file still `registry:component` with no target; every `./*.css` import in a shipped source file has a matching shipped stylesheet, and no stylesheet is shipped unused.
- `npm run build` passes. `npx prettier --check jsrepo.config.ts` passes.
- `npm run lint` reports 46 errors, all pre-existing in `src/**/*.jsx` — eslint runs with `--ext js,jsx`, so it doesn't cover either file this PR changes.

## Out of scope

- Lanyard's `card.glb` / `lanyard.png` still aren't shipped through the registry. They were already excluded before this change, and the component's docs tell users to supply them.
- Lanyard's `dependencyResolution: 'manual'` still leaves its `dependencies` empty — separate issue.


DavidHDev:
Landed in #1053 with contributor authorship preserved. The current registry was regenerated, and a real shadcn 4.18.0 add installation succeeded. Thank you!

Issue #1058, opened 2026-08-31
[BUG]: Documented prop defaults don't match component sources (50 props / 16 components), and 4 variants disagree on 7 props


Agnik47:
### Description

The prop tables on the component docs pages (`propData` in `src/demo/**/*Demo.jsx`) have drifted away from the actual component source. I scanned all 166 components and found **50 props across 16 components where the documented default does not match the real default**, plus **7 props across 5 components where the four variants disagree with each other** (a CONTRIBUTING rule 5 violation).

This matters more than a normal docs typo because these tables are the primary prop reference for the library, and they are also what people paste into AI coding assistants. A wrong default produces code that silently behaves differently from what the docs promised.

### 1. Documented default != actual default

Comparison is against the JS-CSS variant (`src/content/<Category>/<Name>/<Name>.jsx`) vs the `propData` array in `src/demo/<Category>/<Name>Demo.jsx`. Only scalar defaults (string / number / boolean) are listed; arrays, objects and placeholder values like `-` or `undefined` were excluded to avoid false positives.

| Component | Prop | Actual default (source) | Documented default |
|---|---|---|---|
| `ClickSpark` | `sparkColor` | `#fff` | `#f00` |
| `ClickSpark` | `sparkSize` | `10` | `30` |
| `ClickSpark` | `sparkRadius` | `15` | `30` |
| `ClickSpark` | `duration` | `400` | `660` |
| `Ferrofluid` | `scale` | `1.6` | `1` |
| `Ferrofluid` | `sharpness` | `2.5` | `3` |
| `Ferrofluid` | `shimmer` | `1.5` | `1` |
| `Ferrofluid` | `mouseRadius` | `0.35` | `0.3` |
| `GlitchText` | `speed` | `1` | `0.5` (!) |
| `GlitchText` | `enableOnHover` | `true` | `false` (!) |
| `LightRays` | `lightSpread` | `1` | `0.5` |
| `LightRays` | `rayLength` | `2` | `1.0` |
| `LightRays` | `followMouse` | `true` | `false` |
| `LightRays` | `mouseInfluence` | `0.1` | `0.5` |
| `Lightfall` | `speed` | `0.5` | `1` |
| `Lightfall` | `streakCount` | `2` | `8` |
| `Lightfall` | `density` | `0.6` | `1` |
| `Lightfall` | `zoom` | `3` | `2` |
| `Lightfall` | `backgroundGlow` | `0.5` | `1` |
[...]

PR #1060, opened 2026-08-31
fix: correct 48 drifted prop-table defaults and align 4 variant default mismatches

Agnik47:
Fixes the two defects reported in #1058. Split into two commits so you can take one and drop the other.

### 1. `fix(docs)` — 48 prop-table defaults corrected (15 files, docs only)

`propData` in `src/demo/**/*Demo.jsx` had drifted from the components. 48 entries across 16 components documented a value the code does not use. Every corrected value is read from the JS-CSS source, which is the variant the docs page renders.

**No runtime behaviour changes in this commit** — it only touches the documented values.

The clearest one was `TextPressure`, whose table was shifted by a row:

| Prop | Was documented | Actually |
|---|---|---|
| `textColor` | `true` (that's `stroke`'s default) | `#FFFFFF` |
| `strokeColor` | `#FFFFFF` (that's `textColor`'s) | `#FF0000` |
| `className` | `#FF0000` (that's `strokeColor`'s) | `""` |

Others of note: `ClickSpark` had all four of `sparkColor` / `sparkSize` / `sparkRadius` / `duration` wrong; `SideRays` had seven props wrong; `Lightfall` seven; `LightRays` four.

### 2. `fix(variants)` — 3 components where the 4 variants disagreed

CONTRIBUTING rule 5 asks for the variants to stay in sync, but these shipped a different default depending on which variant you installed. In each case 3 of the 4 agree **and** the documented default agrees with that majority, so I corrected the outlier:

| Component | Prop | Change | Was |
|---|---|---|---|
| `GlitchText` | `speed` | JS-CSS `1` → `0.5` | other 3 variants + docs: `0.5` |
| `GlitchText` | `enableOnHover` | JS-CSS `true` → `false` | other 3 variants + docs: `false` |
| `MetaBalls` | `enableTransparency` | JS-CSS `true` → `false` | other 3 variants + docs: `false` |
| `SideRays` | `falloff` | TS-TW `2.0` → `1.6` | other 3 variants: `1.6` |

`GlitchText` was the user-visible one — install JS-CSS and it glitches continuously; install any other variant and it only glitches on hover.

**Preview impact:** the `GlitchText` and `SideRays` demos pass these props explicitly, so nothing changes on screen. The `MetaBalls` demo did *not* pass `enableTransparency` and relied on the JS-CSS default, so flipping it would have made the preview opaque in light mode — the demo now passes `enableTransparency` explicitly, keeping the preview exactly as it is today.

Registry items for the three affected variants (`GlitchText-JS-CSS`, `MetaBalls-JS-CSS`, `SideRays-TS-TW`) are rebuilt to match, following the same pattern as the Hyperspeed and ShapeBlur fixes. Each is a one-line change; no other registry file is touched.

### Deliberately left alone — your call, not mine

Two splits have no majority and no tiebreaker, so I did not guess:

- **`FuzzyText` `fontSize`** — JS variants use `clamp(2rem, 10vw, 10rem)`, TS variants use `clamp(2rem, 8vw, 8rem)`. A 2-2 split, inconsistent since the component was first added. The docs currently say `8vw`.
- **`LiquidChrome` `amplitude` / `frequencyY`** — JS-CSS uses `0.3` / `3`, the other three use `0.5` / `2`, and the docs matched *neither* (`0.6` / `1.5`). I set the docs to the JS-CSS values since that is what the docs page renders, and left all four sources untouched.

Tell me which value you want in each and I'll follow up.

`RippleDistortion`'s `src` is also flagged by my scan, but its table deliberately truncates a long Unsplash URL to `'https://images.unsplash.com/...'`. Left as-is.

### Verification

- `npm run lint` — 44 problems (33 errors, 11 warnings), **identical before and after**; all pre-existing.
- `npx prettier --check` on the touched files — same pass/fail set before and after; no new formatting deviations.
- `npx vite build` — passes.
- Re-ran the audit script after the changes: 0 remaining mismatches apart from the three cases listed above.

### Scope note

My scan could not parse the prop signature of 60 of the 171 components, so this is a lower bound rather than a complete audit. The checker script I offered in #1058 is not included here to keep this PR reviewable — happy to send it separately if you want the class of bug prevented rather than just fixed.

Refs #1058


DavidHDev (approved):
The default corrections and three variant fixes in this PR are sound. I checked the prop-table changes against the source and verified that the three generated registry entries match it. Preserving the MetaBalls preview with an explicit enableTransparency prop is correct.

Validation: TypeScript and the production build pass with these changes; lint findings are unchanged from main.

This is good to merge as a partial fix for #1058. I have prepared a separate local follow-up for the remaining LiquidChrome/FuzzyText splits, additional drift found by a broader AST audit, and a build-time check. The remaining issue work should not block these corrections.

PR #1065, opened 2026-09-05
fix(TextLoop): prevent separator duplication causing overlap (#1062)

### LINKED ISSUE #1062: [BUG]: TextLoop separator overlaps letters because default text already contains the separator
### Describe the issue

On the TextLoop component, one of the ✦ separators occasionally lands on top of a letter instead of sitting cleanly between repetitions.

I noticed this bug while using Firefox on my system. It is visible on the wave shape. You usually have to wait a few seconds for it to appear. Chromium-based browsers often do not show the problem.

### Root cause

The default props are:

text = 'React ✦ Bits'
separator = '✦'

When uppercase is true, the text becomes "REACT ✦ BITS".
The component then appends another separator:

const gap = separator ? `\u00A0${separator}\u00A0` : '\u00A0\u00A0\u00A0';
return `${base}${gap}`;

So the actual string that gets repeated is:

REACT ✦ BITS ✦ REACT ✦ BITS ✦ ...

Having the separator character both inside the phrase and as the joining character breaks the text measurement + textLength / lengthAdjust calculation on the SVG path in stricter browsers (like Firefox).

Image

Image

### Reproduction Link

https://reactbits.dev/text-animations/text-loop

### Steps to reproduce
1. Open https://reactbits.dev/text-animations/text-loop on firefox
2. Wait a few seconds and watch the wave animation — one of the ✦ characters will land on top of a letter (usually the “R” of REACT) 

### Validations

MauryaQbit:
Fixes #1062

## Bug
TextLoop default props used text='React [star] Bits' with separator='[star]'. After uppercase the repeated unit became 'REACT [star] BITS[NBSP][star][NBSP]', i.e. two separators per repetition. The duplicated glyph breaks the textLength / lengthAdjust fit on the SVG path in stricter browsers (Firefox, wave shape), so a [star] can land on top of a letter (usually R of REACT).

## Fix
- Change default text from 'React [star] Bits' to 'React Bits' in all 4 variants, demo DEFAULT_PROPS/prop table, and usage example.
- Harden unit construction: strip any separator occurrences from base, collapse whitespace, trim, then append exactly one '[NBSP][sep][NBSP]' gap. Legacy text containing the separator now converges to the same correct output. Empty separator still falls back to 3x NBSP.
- Applied to: src/content/.../TextLoop.jsx, src/tailwind/.../TextLoop.jsx, src/ts-default/.../TextLoop.tsx, src/ts-tailwind/.../TextLoop.tsx, plus TextLoopDemo.jsx and textLoopCode.js.

## Review notes
- No API change; default rendering changes from 'REACT [star] BITS [star]' to 'REACT BITS [star]' per repetition (single separator, even rhythm).
- Custom text containing the separator is sanitized to one separator per unit per component contract ('Glyph placed between each repetition'). Whitespace is collapsed/trimmed to avoid double spacing at the junction.
- Separator case follows uppercase flag for visual consistency (no-op for glyphs like [star]).
- public/r registry JSONs intentionally untouched (generated artifacts, local jsrepo build shows only CRLF noise).

## Tests
- Repo has no unit test script (package.json: dev/build/lint/format only).
- Verified: node unit-logic script 7/7 PASS (default, legacy default, custom with/without sep inside, text==sep, whitespace collapse, empty sep; all produce exactly 1 sep, 0 when empty).
- Verified: npx tsc --noEmit PASS (0 errors).
- Verified: npx eslint on changed JS files PASS (0 errors); full npm run lint shows 33 pre-existing errors on main, none in TextLoop files.
- Verified: npm run registry:build PASS (684 items). Full vite build timed out locally (>120s, large site) with no related errors.

### DIFF
--- a/src/content/TextAnimations/TextLoop/TextLoop.jsx
+++ b/src/content/TextAnimations/TextLoop/TextLoop.jsx
@@ -45,7 +45,7 @@ const buildPath = (shape, curviness, ribbonWidth) => {
 };
 
 const TextLoop = ({
-  text = 'React ✦ Bits',
+  text = 'React Bits',
   shape = 'wave',
   path,
   speed = 90,
@@ -78,8 +78,22 @@ const TextLoop = ({
   const d = useMemo(() => path || buildPath(shape, curviness, ribbonWidth), [path, shape, curviness, ribbonWidth]);
 
   const unit = useMemo(() => {
-    const base = uppercase ? String(text).toUpperCase() : String(text);
-    const gap = separator ? `\u00A0${separator}\u00A0` : '\u00A0\u00A0\u00A0';
+    let base = String(text ?? '');
+    let sep = separator ?? '';
+    if (uppercase) {
+      base = base.toUpperCase();
+      sep = sep.toUpperCase();
+    }
+    if (sep) {
+      // Strip any separator glyphs already present in the text so each
+      // repetition is joined by exactly one separator. A duplicated glyph
+      // breaks the textLength/lengthAdjust fit on the SVG path in stricter
+      // browsers (e.g. Firefox) and a separator can land on top of a letter.
+      base = base.split(sep).join('').replace(/[\s\u00A0]+/g, ' ').trim();
+    } else {
+      base = base.replace(/[\s\u00A0]+/g, ' ').trim();
+    }
+    const gap = sep ? `\u00A0${sep}\u00A0` : '\u00A0\u00A0\u00A0';
     return `${base}${gap}`;
   }, [text, separator, uppercase]);
 
--- a/src/constants/code/TextAnimations/textLoopCode.js
+++ b/src/constants/code/TextAnimations/textLoopCode.js
@@ -9,7 +9,7 @@ export const textLoop = {
   usage: `import TextLoop from './TextLoop';
 
 <TextLoop
-  text="React ✦ Bits"
+  text="React Bits"
   shape="wave"
   speed={90}
   direction="forward"
--- a/src/demo/TextAnimations/TextLoopDemo.jsx
+++ b/src/demo/TextAnimations/TextLoopDemo.jsx
@@ -19,7 +19,7 @@ import TextLoop from '../../content/TextAnimations/TextLoop/TextLoop';
 import { textLoop } from '../../constants/code/TextAnimations/textLoopCode';
 
 const DEFAULT_PROPS = {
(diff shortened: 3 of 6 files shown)


MauryaQbit:
@rorz can you review this pr

rorz:
Confused as to why I am being mentioned here. This PR / issue has nothing to do with me.

DavidHDev:
@rorz probably a mistake, feel free to disregard it

src/content/TextAnimations/TextLoop/TextLoop.jsx
```diff
@@ -78,8 +78,22 @@ const TextLoop = ({
...
+    if (uppercase) {
+      base = base.toUpperCase();
+      sep = sep.toUpperCase();
+    }
+    if (sep) {
+      // Strip any separator glyphs already present in the text so each
+      // repetition is joined by exactly one separator. A duplicated glyph
+      // breaks the textLength/lengthAdjust fit on the SVG path in stricter
+      // browsers (e.g. Firefox) and a separator can land on top of a letter.
+      base = base.split(sep).join('').replace(/[\s\u00A0]+/g, ' ').trim();
```
DavidHDev, line 92: [P2] Fix the SVG length mismatch instead of deleting separator characters. On this PR, Firefox renders text="A ✦ B" at about 2313.50 SVG units while the loop advances by the path length of 2270.86, so the copies still overlap. Firefox ignores textLength/lengthAdjust on textPath; moving both attributes to each parent text makes the rendered span match 2270.86 (Chromium also stays correct). See Mozilla bug 890692: https://bugzilla.mozilla.org/show_bug.cgi?id=890692. This split/join also changes legitimate content: text="seaside", separator="s", uppercase={false} becomes "eaide", while the accessible label still says "seaside". Preserve the text and address the SVG fitting in all four variants.

DavidHDev (changes requested):
I reproduced the Firefox problem and tested this PR in Firefox and Chromium. The separator removal does not fix the underlying text fitting mismatch, and it changes valid custom text. Please keep the supplied text intact and move textLength={fitLength} and lengthAdjust="spacing" from both textPath elements to their respective parent text elements in all four variants, then regenerate the registry.

With those attributes on text, the original default text works. I verified this alternative in 80 fit cases per browser across the four variants and five shapes, including embedded separators and empty text, and inspected desktop/mobile wave renders. Details below.

PR #1070, opened 2026-09-06
fix(LetterGlitch): smooth colour transitions freeze after one frame

### LINKED ISSUE #1069: [BUG]: LetterGlitch smooth colour transition freezes after one frame
### Describe the issue

I was using LetterGlitch as a page background and noticed that turning `smooth` on and off barely changed anything. Turns out the colour fade does start, but it stops on the very first frame — a letter ends up about 5% of the way to its target colour and stays there until the next glitch tick swaps it out for something else.

So `smooth={true}` is more or less inert right now.

### Root cause

In `handleSmoothTransitions`:

```js
const startRgb = hexToRgb(letter.color);
const endRgb = hexToRgb(letter.targetColor);
if (startRgb && endRgb) {
  letter.color = interpolateColor(startRgb, endRgb, letter.colorProgress);
  needsRedraw = true;
}
```

`letter.color` starts out as a hex string from `getRandomColor()`, but the line inside the `if` overwrites it with `interpolateColor(...)`, which returns `rgb(r, g, b)`. On the next frame `hexToRgb("rgb(46, 77, 62)")` matches neither the shorthand nor the 6-digit regex, so it returns `null`, the `if` never runs again, and the colour is stuck. `colorProgress` still climbs all the way to 1, and `needsRedraw` stays `false`.

`targetColor` isn't affected, since it's always reassigned straight from `getRandomColor()` and stays hex. It's only the start of the interpolation that gets poisoned.

Same code in all four variants:

- `src/content/Backgrounds/LetterGlitch/LetterGlitch.jsx:146`
- `src/tailwind/Backgrounds/LetterGlitch/LetterGlitch.jsx:146`
- `src/ts-default/Backgrounds/LetterGlitch/LetterGlitch.tsx:166`
- `src/ts-tailwind/Backgrounds/LetterGlitch/LetterGlitch.tsx:166`

I think this is easy to miss because the canvas still animates perfectly well — `updateLetters()` swaps characters every `glitchSpeed` ms and calls `drawLetters()` regardless of any of this. The only thing missing is the fade *between* colours, which is subtle enough that it just reads as "smooth doesn't do much".

### Reproduction Link

https://reactbits.dev/backgrounds/letter-glitch

### Steps to reproduce

The two functions don't touch the DOM, so you can watch this happen in plain Node without React or a browser. Both are copied verbatim from the component:

```js
const hexToRgb = hex => {
  const shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
  hex = hex.replace(shorthandRegex, (_m, r, g, b) => r + r + g + g + b + b);
  const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
  return result
    ? { r: parseInt(result[1], 16), g: parseInt(result[2], 16), b: parseInt(result[3], 16) }
    : null;
};

const interpolateColor = (start, end, factor) => {
  const result = {
    r: Math.round(start.r + (end.r - start.r) * factor),
    g: Math.round(start.g + (end.g - start.g) * factor),
    b: Math.round(start.b + (end.b - start.b) * factor)
  };
  return `rgb(${result.r}, ${result.g}, ${result.b})`;
};

// one letter, stepped through handleSmoothTransitions frame by frame
const letter = { color: '#2b4539', targetColor: '#61dca3', colorProgress: 0 };
let redraws = 0;

for (let frame = 1; frame <= 20; frame++) {
  if (letter.colorProgress < 1) {
    letter.colorProgress += 0.05;
    if (letter.colorProgress > 1) letter.colorProgress = 1;

    const startRgb = hexToRgb(letter.color);
    const endRgb = hexToRgb(letter.targetColor);
    if (startRgb && endRgb) {
      letter.color = interpolateColor(startRgb, endRgb, letter.colorProgress);
      redraws++;
    }
    console.log(
      `frame ${String(frame).padStart(2)}  progress ${letter.colorProgress.toFixed(2)}` +
      `  color ${letter.color.padEnd(18)}  ${startRgb ? 'ok' : 'hexToRgb(letter.color) -> null'}`
    );
  }
}

console.log(`\nfinal colour:   ${letter.color}`);
console.log(`expected:       rgb(97, 220, 163)  (= #61dca3)`);
console.log(`frames redrawn: ${redraws} of 20`);
```

Using the component's own default colours (`#2b4539` → `#61dca3`):

```
frame  1  progress 0.05  color rgb(46, 77, 62)     ok
frame  2  progress 0.10  color rgb(46, 77, 62)     hexToRgb(letter.color) -> null
frame  3  progress 0.15  color rgb(46, 77, 62)     hexToRgb(letter.color) -> null
...
frame 20  progress 1.00  color rgb(46, 77, 62)     hexToRgb(letter.color) -> null

final colour:   rgb(46, 77, 62)
expected:       rgb(97, 220, 163)  (= #61dca3)
frames redrawn: 1 of 20
```

One frame out of twenty actually paints.

### Possible fix

Keeping the two interpolation endpoints in hex and letting `letter.color` be render-only fixes it, since the parser is then never handed its own output:

```diff
   letters.current = Array.from({ length: totalLetters }, () => ({
     char: getRandomChar(),
     color: getRandomColor(),
+    startColor: getRandomColor(),
     targetColor: getRandomColor(),
     colorProgress: 1
   }));
```

```diff
       letters.current[index].char = getRandomChar();
+      letters.current[index].startColor = letters.current[index].targetColor;
       letters.current[index].targetColor = getRandomColor();
```

```diff
-        const startRgb = hexToRgb(letter.color);
+        const startRgb = hexToRgb(letter.startColor);
         const endRgb = hexToRgb(letter.targetColor);
```

The other way round would be to have `interpolateColor` return `{r, g, b}` and only format it into a string at the `ctx.fillStyle` assignment — slightly bigger diff, but no extra field. Happy to open a PR with whichever you'd rather have.

### Validations

FelipeKreulich:
Fixes #1069.

`smooth` was effectively doing nothing. The fade started and then stopped on the very next frame, leaving each letter about 5% of the way to its target colour until the next glitch tick swapped it out.

### What was happening

`handleSmoothTransitions` read the fade's starting point from `letter.color`:

```js
const startRgb = hexToRgb(letter.color);
```

but the line just below it overwrites that same field with `interpolateColor(...)`, which returns `rgb(r, g, b)`. From the second frame onwards `hexToRgb` was being handed its own formatted output, matched neither regex, and returned `null` — so the guarded branch never ran again. `colorProgress` still climbed to 1 and `needsRedraw` stayed `false`.

It's easy to miss because the canvas keeps animating regardless: `updateLetters()` swaps characters every `glitchSpeed` ms and calls `drawLetters()` on its own. Only the colour *fade* was missing.

### What I changed

Interpolation now happens in numbers, and the CSS string is built only where it's used, at `ctx.fillStyle`. The parser is never handed its own output, so the failure can't come back.

Each letter now carries:

- `rgb` — what is currently painted
- `fromRgb` / `targetRgb` — the two endpoints of the fade

`interpolateColor` became `mixRgb` (returns `{r, g, b}`) plus `rgbToCss`.

Two things fall out of this that I think are improvements, but say the word if you'd rather I drop them:

1. **A letter picked again mid-fade now continues from the colour on screen** instead of restarting from a value that was never displayed. `updateLetters` sets `fromRgb` to the letter's current `rgb`.
2. **An unparseable entry in `glitchColors` no longer stalls the animation.** Previously it produced an invalid `fillStyle` and a frozen fade; now it falls back to white, which is visible rather than silent. I hit this by accident while testing — the docs page's `?colors=` URL param splits the value incorrectly and passes `#`, `f`, `f` to the component. That looks like a separate demo-side bug, unrelated to this PR; happy to open an issue for it if it's news to you.

### All four variants

`content`, `tailwind`, `ts-default` and `ts-tailwind` all carry the identical change. I checked first that they only differ in their styling layer, so the animation logic stays byte-identical across them. The four `public/r/LetterGlitch-*.json` files were regenerated with `npm run registry:build` rather than hand-edited.

`npm run registry:build` also rewrote `public/r/DecryptedText-{JS,TS}-CSS.json`, which looked like pre-existing drift between that component's source and its registry output. I left those out of this PR since they're unrelated.

### Testing

**Locally, in the docs app**, sampling one letter cell over 90 frames and measuring the distance between consecutive colours:

| | distinct colours | gradual steps | hard jumps |
|---|---|---|---|
| `smooth` on | 42 | 40 | 0 |
| `smooth` off | 7 | 0 | 6 |

Before the fix these two columns were near-identical, which is the bug. Now `smooth` on walks through intermediate tones, and `smooth` off jumps straight between palette colours — which is what each mode should do.

Repeated at a 360×640 container to cover the mobile resize path, since `resizeCanvas` → `initializeLetters` is where the new per-letter state is built from scratch: 43 distinct colours, 42 gradual steps, 0 hard jumps. Browser console is clean — no errors, and no new warnings (the only ones present are the pre-existing React Router v7 future-flag notices).

**Type checking:** both `.tsx` variants pass `tsc --noEmit --strict`.

**Lint:** `npx eslint` on the two `.jsx` variants reports nothing, before and after the change. (Repo-wide `npm run lint` reports 44 pre-existing problems in other components, none in these files.)

There's also a dependency-free Node reproduction of the original bug in #1069 if it's useful for confirming the diagnosis.


DavidHDev (approved):
Reviewed all four source variants and verified that the four generated registry entries exactly match their source. The numeric interpolation fixes #1069 and correctly restarts an interrupted fade from the currently painted color.

Validation: deterministic frame checks on every variant reach the target after 20 fade steps, continue correctly when interrupted, and switch immediately with smooth=false. TypeScript and the production build pass with the change applied; lint has no new findings. No blocking findings.

PR #1083, opened 2026-09-15
fix(PixelSwap): add missing preview videos

EnderRomantice:
## Summary

The PixelSwap entry already references `/assets/video/pixelswap.webm`, but the preview files are missing. As a result, its card in the component index falls back to the component-name placeholder instead of showing the animation.

Add `pixelswap.webm` and `pixelswap.mp4` so the existing card player can show the recorded forward and reverse pixel-swap transition. The current metadata and dual-format source selection automatically pick up these assets.

## Video assets

Both clips contain the full 5.37-second recording, resized to the existing preview format of **768 × 480 at 30 fps**, without audio.

| File | Encoding | Size |
| --- | --- | --- |
| `public/assets/video/pixelswap.webm` | VP9, yuv420p | 125 KiB |
| `public/assets/video/pixelswap.mp4` | H.264, yuv420p, faststart | 136 KiB |

https://github.com/user-attachments/assets/d180ad37-fc6f-4b2e-890e-beae55b01e97

## Validation

- Decoded both files completely with FFmpeg without errors and checked their metadata with FFprobe.
- Verified the PixelSwap card loads the WebM preview in the local component index on desktop and at a 390 × 844 mobile viewport.
- Verified the H.264 MP4 plays in the browser.
- Checked the browser console: no errors; only existing React Router future-flag warnings.


DavidHDev:
thank you!

PR #1086, opened 2026-09-16
fix(BackgroundStudio): keep preview controls clickable over background content

a10yuge:
## Bug

On the Background Studio (`/tools/background-studio`), the preview toolbar controls stop responding to clicks while certain backgrounds are selected. With **Image Trail** active, the Previous/Next background arrows in the top-right corner of the preview do nothing.

Steps to reproduce:

1. Open https://reactbits.dev/tools/background-studio?bg=image-trail
2. Click the `‹` / `›` buttons in the top-right corner of the preview
3. Nothing happens — the background never changes

The Left/Right keyboard shortcuts still work, which makes this easy to miss.

## Cause

The rendered background was never isolated into its own stacking context.

`BackgroundRenderer` mounts the background into an absolutely positioned `Box` with `z-index: auto`, and backgrounds are standalone components that set their own z-index — `ImageTrail`'s root `.content` uses `z-index: 100`, while the studio's own overlays use `zIndex={99}`. Because the wrapper does not create a stacking context, that `100` was resolved in the page's stacking context instead of inside the preview, so the background's root element painted over the toolbar and swallowed the pointer events.

`document.elementFromPoint()` on the button centers returns `div.content`, and Playwright's hit-test refuses the click with `<div class="content">…</div> from <div>…</div> subtree intercepts pointer events`.

Simply raising the toolbar's z-index would only be a band-aid: ImageTrail assigns its images an incrementing z-index while the trail is active, so any fixed value can eventually be exceeded.

## Fix

`isolation: isolate` on the background wrapper. Every z-index a background declares now stays inside that layer, so the studio UI is always on top no matter what the rendered background does.

## Verification

`elementFromPoint` on each control's center, before → after:

| control | before | after |
| --- | --- | --- |
| Previous / Next background | `div.content` | the button |
| Image / 10s Video | `div.content` | the button |
| mobile Controls button (bottom-right) | `div.content` | the button |

Five controls were affected in total, not only the two arrows.

Also re-checked:

* Left/Right keyboard shortcuts still switch backgrounds
* The other backgrounds that render outside the preview area (e.g. SplashCursor, which uses `position: fixed`) are pixel-identical before/after
* The docs demos are untouched, since only the studio's render wrapper changed
* `npm run lint` on the changed file and `npm run check:prop-docs` both pass

## Unrelated note

While testing I ran into a separate crash: switching from Iridescence to Laser Flow with the arrows blanks the entire page (`Cannot read properties of null (reading '1')`, thrown from ogl's `hexToRGB`). `BackgroundRenderer` keeps the previously loaded component in state while already handing it the new background's props, so Iridescence receives Laser Flow's `color` string where it expects an array. Happy to file that separately if useful — it is not addressed here.


src/tools/background-studio/BackgroundStudio.jsx
```diff
@@ -121,6 +121,12 @@ const BackgroundRenderer = React.memo(({ background, props, renderKey }) => {
 
   const wrapperStyle = background.wrapperStyle || {};
 
+  // Backgrounds are standalone components that manage their own z-index, and some of them
```
DavidHDev, line 124: No need for all these comments, please remove them so I can merge this.

PR #1088, opened 2026-09-17
Fix ASCIIText in light mode

Brisk4t:
### ASCIIText used to change its characters and drop the color animation in light mode 

**Before:**
image

Needed to remove the base theme dependence and invert the colors after the render instead. Idk if its the best way but its the least intrusive.

**After:**
image


DavidHDev:
Good stuff, but let's not add overly verbose comments to the code, can you please remove them, or keep it a simple one line comment?

DavidHDev (approved):

